PR #1, opened 2020-09-15
Complete GitHub migration for API

marcellp:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,165 @@
+name: Build and Deploy the API
+on:
```
ivababukova: make it `on: [push, pull_request]` instead of this, so that builds trigger for branches as well

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,165 @@
...
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64
+          tags: ${{ format('{0}/{1}:{2}', steps.login-ecr.outputs.registry, steps.ref.outputs.repo-name, steps.ref.outputs.tag) }}
+          push: false
+
+      - id: push
+        name: Push docker image to ECR
+        run: |-
```
ivababukova, line 92: make it push the image only if there is a push to master branch

.github/workflows/ci.yaml
marcellp: Do we /need/ this though? I thought about this, and the rationale was the times we really care about the CI's status is when

(a) a pull request is out for a feature, in which case each subsequent push on a branch with an open pull request will trigger the CI pipeline. We are only opening pull requests to master.
(b) we are merging stuff into master, in which case there is a push event to master that is triggered.

I think restricting the `push` trigger to `master` makes sense in this light -- there is no use running CI pipelines for dev branches that are not ready for review. If you think the CI should run on MRs for all branches (not just `master`), in case we ever decide to do a different branching model, I can make it:

    push:
        branches:
          - master
    pull_request:

ivababukova: I think it is still useful to have it, because then it is really easy to see in which build what went wrong. Also, it is a standard thing to run the build for every push to a branch so I would prefer to stick with the standard guidance on this. If we see in the future that it is too much noise, of course we can revisit.

marcellp: Sure, changed.

.github/workflows/ci.yaml
marcellp: Fixed.

ivababukova (approved):

PR #3, opened 2020-09-15
Use Redis on the API

marcellp:


src/cache/index.js
```diff
@@ -1,103 +1,133 @@
...
       // unstringify the data
       result = JSON.parse(result);
       return result;
     } catch (error) {
-      logger.error('Failed to get item from cache', error);
-      logger.trace(error);
-      this.redisClient.slave.checkStatus();
+      const message = `redis:reader Cannot GET from ${key} as an error occurred: ${error.message}`;
+      logger.error(message);
+      throw new Error(message);
```
ivababukova, line 109: why are we throwing the error now?

src/cache/cache-instances.js
```diff
@@ -1,25 +0,0 @@
...
-    slaves: [
-      {
-        host: 'localhost',
-        port: '6379',
-      },
-    ],
-  };
-};
-
-module.exports = getInstances;
```
ivababukova, line 25: interesting, it seems like this file was never used?

src/cache/cache-list.js
```diff
@@ -1,39 +0,0 @@
...
-    ReplicationGroupId: `biomage-redis-${clusterEnv}`
-  }).promise();
-
-
-  // biomage - redis - production
-}
-
-
-module.exports = {
-  'eu-west-1': {
```
ivababukova, line 2: I wonder how did that ever work if it is saying eu-west-1 and we used to be in eu-west-2

src/cache/index.js
marcellp: There is a `try...catch` block in `get` that uses `_redisGet` internally, so I thought it makes more sense to throw an exception rather than returning `null`.

marcellp: It also simplifies the code because there is no need to run an `if(result)` check inside `get()`.

src/cache/cache-instances.js
marcellp: It was used, but the file it was used was deleted in this release too. The entire dependency chain was very deep to build configuration object that was not really necessary, so everything got abstracted into `generate-config.js`.

src/cache/cache-list.js
marcellp: I changed that in a previous commit I'm pretty sure when I was migrating to GitHub. It says `eu-west-2` on GiLab, see [here](https://gitlab.com/biomage/api/-/blob/master/src/cache/cache-list.js).

src/cache/cache-instances.js
marcellp: For reference, it goes: `cache/index.js` > `cache/sanitize.js` (via `sanitiseGlobalConfiguration`), which calls `getInstances`.

ivababukova: I just looked at all of them and at your implementation and I understand now. Your changes are so much better! The previous code was so hardcoded and poorly written that I wonder how it managed to make it to master in that state

src/cache/index.js
ivababukova: yup, fair enough. Do we not need to handle that error from somewhere though? Like, we don't want the API to blow up if the key is not in Redis, right?

marcellp: It's caught as we discussed today.

tests/api/event-services/work-request.test.js
```diff
@@ -1,73 +1,27 @@
...
-    emit: emitSpy,
-  };
-  toSpy = jest.fn(() => mockEmit);
+jest.mock('../../../src/cache');
 
-  io = {
-    to: toSpy,
-    emit: emitSpy,
-  };
-};
```
marcellp, line 53: This mock is replaced by `socket.io-mock`, which provides a proper interface for these.

tests/api/event-services/work-request.test.js
```diff
@@ -1,73 +1,27 @@
...
-let toSpy;
-
-const setIoMock = () => {
-  emitSpy = jest.fn();
-
-  const mockEmit = {
-    emit: emitSpy,
-  };
-  toSpy = jest.fn(() => mockEmit);
+jest.mock('../../../src/cache');
```
marcellp, line 7: The new mock is replacing the ugly harcoded one we used to have.

tests/api/routes/health.test.js
```diff
@@ -1,12 +1,15 @@
 const express = require('express');
 const request = require('supertest');
 const expressLoader = require('../../../src/loaders/express');
+const CacheSingleton = require('../../../src/cache');
```
marcellp, line 4: We need to mock cache because `health` checks for cache connectivity.

tests/cache/cache.test.js
```diff
@@ -1,83 +0,0 @@
...
-  set: jest.fn(),
-})));
-
-jest.mock('../../src/cache/aws-utilities', () => jest.fn((value) => new Promise((resolve, reject) => {
-  if (value === 'placement/availability-zone') {
-    resolve('eu-west-1b');
-  } else {
-    reject(new Error('error'));
-  }
-})));
```
marcellp, line 43: All of this is replaced by the real deal (e.g. with `lru`), or a mocked library instead of a manual mock.

See `index-l1-hot.js`.

tests/cache/index-l1-hot.test.js
```diff
@@ -0,0 +1,183 @@
...
+  // ES constructor
+
+  // eslint-disable-next-line func-names
+  return function (args) {
+    const object = new Redis({ ...args, lazyConnect: false });
+    object.status = 'ready';
+
+    return object;
+  };
+});
```
marcellp, line 30: This is a mock that generates a mocked Redis client using `ioredis-mock`. There is a weird bug going on in it that requires this complicated shimming logic.

https://github.com/stipsan/ioredis-mock/issues/568

tests/utils/cache-request.test.js
```diff
@@ -1,47 +1,45 @@
-const { cacheSetResponse, cacheGetRequest, CacheMissError } = require('../../src/utils/cache-request');
-const cache = require('../../src/cache');
-
-jest.mock('../../src/cache', () => ({
-  get: jest.fn((key) => {
-    if (key === '7b18513d240c9ad9dfd003a39a6be9fb') return { result: 'valueInL1' };
-    return null;
-  }),
-  set: jest.fn(),
-}));
```
marcellp, line 10: Again, the weird stub is gone.

tests/api/route-services/work-response.test.js
```diff
@@ -1,90 +1,80 @@
...
-        timeout: '2001-01-01T00:00:00Z',
-        body: {
-          name: 'GetEmbedding',
-          type: 'pca',
-        },
-      },
-      response: {
-        error: false,
-        cacheable: true,
-      },
```
marcellp, line 87: These replicas are gone.

tests/api/route-services/work-response.test.js
```diff
@@ -1,90 +1,80 @@
...
+      'reconnection delay': 0,
+      'reopen delay': 0,
+      'force new connection': true,
+      transports: ['websocket'],
+    });
+
+    client.on('connect', () => {
+      responseFormat.request.socketId = client.id;
+      done();
+    });
```
marcellp: The old mocking logic was a mess, so I did this -- it now uses the real `socket.io`. A client is instantiated at every run and connects to the server running alongside in the test. This way we can get a better test of this behavior that is also easier to understand.

tests/api/route-services/work-response.test.js
```diff
@@ -124,35 +105,22 @@ describe('tests for the work-response service', () => {
...
 
-    w.handleResponse().then(() => {
-      expect(toSpy).toHaveBeenCalledTimes(1);
-      expect(toSpy).toHaveBeenCalledWith('mysocketid');
-      expect(emitSpy).toHaveBeenCalledTimes(1);
-      expect(emitSpy).toHaveBeenCalledWith('WorkResponse-55', expectedResponse);
-      return done();
+    client.on(`WorkResponse-${workResponse.request.uuid}`, (res) => {
+      expect(res).toEqual(expectedResponse);
+      done();
```
marcellp, line 118: Instead of the many calls to `emitSpy`, `toSpy`, etc., we can just directly check if the client got the message, and if it was the appropriate message to the appropriate channel, which is much more realistic.

tests/api/route-services/work-response.test.js
```diff
@@ -173,93 +141,71 @@ describe('tests for the work-response service', () => {
...
-      expect(emitSpy).toHaveBeenCalledWith('WorkResponse-55', expectedResponse);
-      expect(getObjectSpy).toHaveBeenCalledTimes(1);
-      expect(getObjectSpy).toHaveBeenCalledWith(
-        {
-          Bucket: 'worker-results-staging',
-          Key: '509520fe-d329-437d-8752-b5868ad59425/48637d30-a88d-481e-b8cc-5eedd7e3af1c',
-          ResponseContentType: 'application/json',
-          ResponseContentEncoding: 'utf-8',
-        },
-      );
```
marcellp, line 237: Again, instead of these old spies, we can just assert directly inside the actual mock, no need to forward to a `getObjectSpy`.

tests/api/route-services/work-response.test.js
```diff
@@ -270,103 +216,86 @@ describe('tests for the work-response service', () => {
...
-        },
-      ], pagination);
-      return done();
+      const expectedBody = JSON.parse(RESPONSE_BODY);
+      expectedBody.total = expectedBody.rows.length;
+      expectedBody.rows = expectedBody.rows.slice(0, pagination.limit);
+      expectedResponse.results[0].body = JSON.stringify(expectedBody);
+
+      expect(res).toEqual(expectedResponse);
+      done();
```
marcellp, line 287: In the old tests handlePagination was mocked out so it wasn't really testing anything. This is now fixed.

src/api/route-services/work-response.js
```diff
@@ -106,15 +103,16 @@ class WorkResponseService {
...
     } catch (e) {
       logger.error('Error trying to cache or paginate data: ', e);
+      return;
     }
+
+    if (Date.parse(timeout) > Date.now()) {
+      this.io.to(socketId).emit(`WorkResponse-${uuid}`, responseForClient);
+    }
+
+    logger.log('Work response sent out.');
```
ivababukova: even if the timeout has passed, the logs will still say `Work response sent out.` which will be wrong. To fix this, better put the logger.log line inside the if statement and have an else statement that is a log message saying that timeout has expired?

tests/api/event-services/work-request.test.js
ivababukova: nice!

tests/api/route-services/work-response.test.js
ivababukova: that is much better indeed, great stuff :)

tests/api/route-services/work-response.test.js
```diff
@@ -1,90 +1,80 @@
...
+    if (client.connected) {
+      client.disconnect();
+    }
   });
 
   it('Throws during validation if invalid data is supplied', async (done) => {
     try {
-      // eslint-disable-next-line no-unused-vars
-      const w = new WorkResponseService(null, {});
+      // eslint-disable-next-line no-new
```
ivababukova, line 70: do we still need this eslint-disable there?

marcellp: yes, there is a "don't use constructor for side effects" warning otherwise

tests/cache/index-l1-hot.test.js
ivababukova: hmmm, interesting

tests/cache/index-l1-hot.test.js
```diff
@@ -0,0 +1,181 @@
...
+  });
+  it('will check cache status if get fails', async () => {
+    cache.redisClient.slave.isHealthy = false;
+    const result = await cache.get('existInRedis');
+    expect(result).toBe(null);
+    expect(cache.redisClient.slave.checkStatus).toHaveBeenCalled();
+    expect(logger.error).toHaveBeenCalledTimes(1);
+  });
+  it('will not set the value in cache if master is not healthy', async () => {
+    cache.redisClient.master.isHealthy = false;
```
ivababukova: these will need to be deleted

src/api/route-services/work-response.js
marcellp: good point, fixed

tests/cache/index-l1-hot.test.js
marcellp: done

ivababukova (approved):

PR #21, opened 2020-10-29
Add support for sandbox ID to API

marcellp:


.github/workflows/ci.yaml
```diff
@@ -156,6 +157,8 @@ jobs:
           access-token: ${{ secrets.API_TOKEN_GITHUB }}
           owner: biomage-ltd
           repo: iac
+          enforce_admins: false
+          retries: 8
```
ivababukova, line 161: why do we need this and why 8?

marcellp: It's just an exponential backoff in case something is preventing the thing from enforcing the controls (e.g. a race condition with another repo trying to write to iac). No specific reason for 8.

ivababukova (approved):

PR #25, opened 2020-11-02
Create staging deployment for master and add sandbox ID

marcellp:


.github/workflows/ci.yaml
```diff
@@ -138,21 +138,33 @@ jobs:
           DEPLOYMENT_NAME=$(echo $GITHUB_REPOSITORY | awk -F '/' '{print $2}')
           echo "::set-output name=deployment-name::$DEPLOYMENT_NAME"
           echo "::set-output name=ref-dashed::$REF_DASHED"
+
+          if [ "${{ matrix.environment }}" = "production" ]; then
+            SANDBOX_ID="default"
```
ivababukova, line 143: This is not a great name. Rename `SANDBOX_ID` to something that doesn't contain `sandbox`, because this is applied not only for sandbox, but for prod as well. Also, not sure about the `ID` part either - it doesn't sound like an id, but more like an environment name

ivababukova: maybe a better name would be `NAMESPACE_SUFFIX`?

marcellp: To me, `sandbox` means an individual isolated environment that can run in a singular cluster. The reason for choosing this was because all the other names are taken by other resources (deployment is a k8s resource, release is a helm resource, namespace is a k8s resource, etc., environment is what we use for staging/production/development). I am happy to re-name it, although at this point it occurs in quite a lot of places over multiple MRs.

As for the `ID` part, it will be an ID when staging environments are created. They will be assigned a SHA1 string, `default` is just the normal one that always runs a copy of production.

ivababukova (approved):

PR #27, opened 2020-11-03
Extract worker images from staging/production deployment files

marcellp:


.secrets.baseline
```diff
@@ -2,7 +2,7 @@
   "custom_plugin_paths": [],
   "exclude": {
     "files": "package-lock.json",
-    "lines": null
+    "lines": ".*integrity.*"
```
ivababukova, line 5: what does that do?

marcellp: there was a bug in the secrets, and it gave a false positive now that i added a new package. i added now the file `package-lock.json` and the lines `integrity` (as appearing in `package-lock.json` with the long string) are excluded from the secrets search

ivababukova (approved):

PR #29, opened 2020-11-06
Add staging support to the API

marcellp:


src/api/routes/health.js
```diff
@@ -5,6 +5,7 @@ const CacheSingleton = require('../../cache');
 module.exports = {
   'health#check': (req, res) => {
     res.json({
+      hi: 'there',
```
ivababukova: we don't need this

ivababukova (approved):

PR #40, opened 2021-01-20
In WorkSchemaGetEmbedding, replace 'type' property with 'config'

iguanaonmystack:
In WorkSchemaGetEmbedding, replace 'type' property with 'config', an abitrary object.

# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-445

https://ui-biomage-445.scp-staging.biomage.net/

I'm not sure this is done but I wanted to make sure I haven't misunderstood the task. This would allow any object as defined by convention between the UI and worker as long as it has a name and 'config' member, which would contain the type, UI settings and so on. As mentioned in the ticket, making this any more rigidly defined would slow down faster iteration of both components.

# Changes
### Code changes

* Swagger yaml spec updated

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:
- [x] After the PR is approved, the `unstage` script in here: https://github.com/biomage-ltd/biomage-utils is executed. This script cleans up your deployment to staging

### Optional

### DIFF
--- a/src/specs/api.yaml
+++ b/src/specs/api.yaml
@@ -458,16 +458,12 @@ components:
           type: string
           example: GetEmbedding
           description: The name of this work -- must be `GetEmbedding`.
-        type:
-          type: string
-          enum:
-            - pca
-            - tsne
-            - umap
-          description: The type of embedding to compute.
+        config:
+          type: object
+          description: The configuration settings to use for this embedding. This will be used both to configure the embedding and to cache the results.
       required:
         - name
-        - type
+        - config
     WorkSchemaListGenes:
       title: WorkSchemaListGenes
       type: object
--- a/src/specs/api.yaml
+++ b/src/specs/api.yaml
@@ -460,7 +460,7 @@ components:
           description: The name of this work -- must be `GetEmbedding`.
         config:
           type: object
-          description: The configuration settings to use for this embedding. This will be used both to configure the embedding and to cache the results.
+          description: The configuration settings to use for calculation of this embedding.
       required:
         - name
     WorkSchemaListGenes:
--- a/src/specs/api.yaml
+++ b/src/specs/api.yaml
@@ -463,7 +463,6 @@ components:
           description: The configuration settings to use for this embedding. This will be used both to configure the embedding and to cache the results.
       required:
         - name
-        - config
     WorkSchemaListGenes:
       title: WorkSchemaListGenes
       type: object
--- a/src/specs/api.yaml
+++ b/src/specs/api.yaml
@@ -463,6 +463,15 @@ components:
           description: The configuration settings to use for calculation of this embedding.
       required:
         - name
+    WorkResponseGetEmbedding:
+      type: array
+      description: 'The response schema for the `GetEmbedding` task.'
+      items:
+        type: array
+        items:
+          type: number
+          minItems: 2
+          maxItems: 2
     WorkSchemaListGenes:
       title: WorkSchemaListGenes
       type: object
--- a/src/specs/api.yaml
+++ b/src/specs/api.yaml
@@ -467,7 +467,8 @@ components:
             - pca
             - tsne
             - umap
-          description: The type of embedding to compute (deprecated; use config).
+          description: The type of embedding to compute.
+          deprecated: true
       required:
         - name
     WorkResponseGetEmbedding:
--- a/src/specs/api.yaml
+++ b/src/specs/api.yaml
@@ -461,6 +461,13 @@ components:
         config:
           type: object
           description: The configuration settings to use for calculation of this embedding.
+        type:
+          type: string
+          enum:
+            - pca
+            - tsne
+            - umap
+          description: The type of embedding to compute (deprecated; use config).
       required:
         - name
     WorkResponseGetEmbedding:
--- a/src/specs/api.yaml
+++ b/src/specs/api.yaml
@@ -457,6 +457,7 @@ components:
         name:
           type: string
           example: GetEmbedding
+          pattern: GetEmbedding
           description: The name of this work -- must be `GetEmbedding`.
         config:
           type: object
--- a/tests/api/event-services/work-request.test.js
+++ b/tests/api/event-services/work-request.test.js
@@ -29,7 +29,7 @@ describe('handleWorkRequest', () => {
       socketId: '6789',
       experimentId: 'my-experiment',
       timeout: '2001-01-01T00:00:00Z',
-      body: { name: 'GetEmbedding', type: 'tsne' },
+      body: { name: 'GetEmbedding', config: {} },
     };
 
     try {
--- a/tests/api/general-services/work-submit.test.js
+++ b/tests/api/general-services/work-submit.test.js
@@ -11,7 +11,7 @@ describe('tests for the work-submit service', () => {
       socketId: '6789',
       experimentId: 'my-experiment',
       timeout: '2099-01-01T00:00:00Z',
-      body: { name: 'GetEmbedding', type: 'pca' },
+      body: { name: 'GetEmbedding', config: { type: 'pca' } },
     };
 
     AWSMock.setSDKInstance(AWS);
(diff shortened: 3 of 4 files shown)


src/specs/api.yaml
```diff
@@ -458,16 +458,11 @@ components:
...
-        type:
-          type: string
-          enum:
-            - pca
-            - tsne
-            - umap
-          description: The type of embedding to compute.
+        config:
+          type: object
+          description: The configuration settings to use for this embedding. This will be used both to configure the embedding and to cache the results.
```
ivababukova: I would rephrase this description as `The configuration settings to use for calculation of this embedding.` I wouldn't have the `This will be used both to configure the embedding and to cache the results.` part because the caching applies to all the tasks, not only to the embedding

src/specs/api.yaml
```diff
@@ -458,16 +458,11 @@ components:
           type: string
           example: GetEmbedding
           description: The name of this work -- must be `GetEmbedding`.
```
ivababukova, line 461: I see that there is no schema defined for the work response for GetEmbedding -- can you add one too? This will make our job easier in the future when we decide to enable the validation of responses received from the worker.

marcellp:
@ivababukova -- my understanding is that this will break the worker unless we merge the other changes in first.

@iguanaonmystack -- could you deploy this to staging to make sure things work as intended?

iguanaonmystack:
@marcellp I've deployed it here: https://ui-98rxkh16615sjk74pbu3y85j9a.scp-staging.biomage.net/experiments/5e959f9c9f4b120771249001/data-exploration

iguanaonmystack:
I wasn't sure which bit Marcel indicated might break, so I've added the WorkResponseGetEmbedding anyway. I'm not sure what tests it needs at this point if we're waiting on other changes first?

ivababukova:
@marcellp I don't think it will break the worker, why would it? Is the worker using the `type` parameter in the `GetEmbedding` task at the moment?

ivababukova:
I took a look into the worker code just now, Marcell is right and this will break the worker, see: https://github.com/biomage-ltd/worker/blob/master/python/src/tasks/embedding.py#L55 and also see the UI: https://github.com/biomage-ltd/ui/blob/master/src/redux/actions/embedding/loadEmbedding.js#L22 At the moment we are using `type`, so we have to keep using type. In order to be able to merge this now without breaking anything, you should undo the changes that remove "type" from the schema and only have `config` added as an additional optional parameter.

iguanaonmystack:
I hadn't realised that it would be a problem if the existing code was sending an additional property the schema didn't recognise. I've added `type` back in now

ivababukova:
The problem is not with the additional property, but with the deletion of the property`type`

marcellp:
@ivababukova -- I thought it would be fine, but my concern was the `type` parameter in the worker, yes.

@iguanaonmystack shares my understanding, as far as I know OpenAPI schemas in general allow additional fields to be declared (such as `type` in this case) as long as they do not interfere with something that actually /is/ in the spec (i.e. there is a field `settings` that is declared to be an object in the schema, it *must* be an object, and not a string).

That said, I wasn't entirely sure, so I wanted to see it work in staging before merging.

marcellp (dismissed):
I think @iguanaonmystack was correct, but the system is actually broken in staging. We have the following error:

```
[2021-01-22T11:06:06.673Z] Error while processing WorkRequest event: Error: Error: bodymatches more than one entry in a oneOf.
    at validateRequest (/app/src/utils/schema-validator.js:16:11)
    at handleWorkRequest (/app/src/api/event-services/work-request.js:28:7)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
[2021-01-22T11:06:06.673Z] Trace: Error: Error: bodymatches more than one entry in a oneOf.
    at validateRequest (/app/src/utils/schema-validator.js:16:11)
    at handleWorkRequest (/app/src/api/event-services/work-request.js:28:7)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at Object.trace (/app/src/utils/logging.js:15:13)
    at Socket.<anonymous> (/app/src/api/events/index.js:13:14)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Seems like one of the new schemas is not well-described. Could you look into why this is happening?

src/specs/api.yaml
```diff
@@ -458,16 +458,27 @@ components:
...
+          type: object
+          description: The configuration settings to use for calculation of this embedding.
         type:
           type: string
           enum:
             - pca
             - tsne
             - umap
-          description: The type of embedding to compute.
+          description: The type of embedding to compute (deprecated; use config).
```
marcellp: You can add a deprecated tag for this in Stoplight Studio, it's a good idea because then we could check later for deprecated stuff being used.

src/specs/api.yaml
```diff
@@ -458,16 +458,27 @@ components:
           type: string
           example: GetEmbedding
           description: The name of this work -- must be `GetEmbedding`.
```
marcellp: My understanding is that the API is failing because you did not set up a pattern for the "type" string like it is done [here](https://github.com/biomage-ltd/api/blob/02d51c45bf6698128832c97558cc7b6986fe082c/src/specs/api.yaml#L489) -- for this schema, only `name` is required, and name does not have a pattern. Therefore, any and all work tasks will match with this, as well as with their respective request schemas.

marcellp:
In general, I would strongly encourage using Stoplight Studio for this -- the OpenAPI schema is really complicated and we have always found things got messed up when we manually edited the schema.

ivababukova (dismissed):

ivababukova:
> I think @iguanaonmystack was correct, but the system is actually broken in staging. We have the following error:
> 
> ```
> [2021-01-22T11:06:06.673Z] Error while processing WorkRequest event: Error: Error: bodymatches more than one entry in a oneOf.
>     at validateRequest (/app/src/utils/schema-validator.js:16:11)
>     at handleWorkRequest (/app/src/api/event-services/work-request.js:28:7)
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (internal/process/task_queues.js:97:5)
> [2021-01-22T11:06:06.673Z] Trace: Error: Error: bodymatches more than one entry in a oneOf.
>     at validateRequest (/app/src/utils/schema-validator.js:16:11)
>     at handleWorkRequest (/app/src/api/event-services/work-request.js:28:7)
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (internal/process/task_queues.js:97:5)
>     at Object.trace (/app/src/utils/logging.js:15:13)
>     at Socket.<anonymous> (/app/src/api/events/index.js:13:14)
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (internal/process/task_queues.js:97:5)
> ```
> 
> Seems like one of the new schemas is not well-described. Could you look into why this is happening?

@iguanaonmystack how is that going? Is the cause of the error found and fixed?

iguanaonmystack:
I'd given Stoplight a go but it removed the single quotes from a whole bunch of strings in the yaml, which made everything difficult to diff. Does this happen for anyone else; are the extra changes it puts in just generally ignored?

iguanaonmystack:
@marcellp Where do I find those logs in staging? I've been poking around in Kontena Lens and the best I've managed is a shell on the node. I seem to recall when you demoed it you just clicked a button somewhere and got a terminal tailing the logs/stdout somehow?

iguanaonmystack:
I've tried to re-deploy the staging site but I'm getting a permissions error: https://github.com/biomage-ltd/iac/actions/runs/509079936

iguanaonmystack:
Removing the staging site and re-creating it (https://ui-biomage-445.scp-staging.biomage.net/experiments/5e959f9c9f4b120771249001/data-exploration) seems to have worked. This looks okay to me now?

ivababukova (approved):

PR #45, opened 2021-02-11
API can persist data-processing progress

marcellp:
# Background
#### Link to issue
https://biomage.atlassian.net/browse/BIOMAGE-500
https://biomage.atlassian.net/browse/BIOMAGE-534

https://ui-gk4wi78my831kkcxbf13nz8gkh.scp-staging.biomage.net/

N/A

Joined the two tickets together are they both involved some refactoring of the API code.

# Changes
### Code changes
* Refactored the API schemas to be in separate files for easier readability of diffs in merge requests.
* Added an additional schema for the `meta` part of the Redux store.
* Persisting processing steps can then be done by the existing Redux action for saving the actual steps' settings.

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/utils/schema-validator.js
```diff
@@ -1,15 +1,39 @@
...
+        statusText: 'OK',
+        url,
+        headers,
+        text: JSON.stringify(data),
+        data: JSON.stringify(data),
+        body: data,
+        obj: data,
+      };
+    },
+  });
```
marcellp: Since `swagger-js` does not support local files, I had to write a mock HTTP client that will open local files, read their contents, and return an OK response. This ensures we can split our files and still validate work requests as necessary.

src/api/route-services/experiment.js
```diff
@@ -116,7 +124,9 @@ class ExperimentService {
     };
 
     const result = await dynamodb.updateItem(params).promise();
-    return result;
+    const prettyData = convertToJsObject(result.Attributes);
+
+    return prettyData;
```
marcellp, line 126: Return the pretty data instead of the weird DynamoDB convention when we do a PUT, as the REST conventions dictate.

src/specs/models/processing-config-bodies/ProcessingConfigMeta.v1.yaml
```diff
@@ -0,0 +1,13 @@
...
+properties:
+  complete:
+    type: boolean
+  stepsDone:
+    type: array
+    items:
+      type: string
+required:
+  - complete
+  - stepsDone
```
marcellp, line 13: New spec for metadata such as completion and steps done.

marcellp:
Staging available at https://ui-gk4wi78my831kkcxbf13nz8gkh.scp-staging.biomage.net/.

tests/api/routes/experiment.test.js
```diff
@@ -134,30 +134,10 @@ describe('tests for experiment route', () => {
...
-          },
-        },
-      },
-    };
-
     request(app)
       .put('/v1/experiments/someId/processingConfig')
       .send(newData)
-      .expect(200, result)
+      .expect(200)
```
aerlaut, line 140: You don't check for the validity of the resulting data?

src/utils/schema-validator.js
aerlaut: By "local files" do you mean local $refs?

aerlaut: Is there a reason to having the keys `text` and `data` with the same value?

src/utils/schema-validator.js
```diff
@@ -1,15 +1,39 @@
...
+      const data = yaml.safeLoad(fs.readFileSync(specPath), 'utf8');
 
+      return {
+        ok: true,
+        status: 200,
+        statusText: 'OK',
+        url,
+        headers,
+        text: JSON.stringify(data),
+        data: JSON.stringify(data),
```
aerlaut, line 26: Is there a reason to having `text` and `data` with the same value?

src/api/route-services/__mocks__/experiment.js
```diff
@@ -52,21 +52,30 @@ const mockGetProcessingConfig = jest.fn(() => new Promise((resolve) => {
 const mockUpdateProcessingConfig = jest.fn(
   () => new Promise((resolve) => {
     resolve({
-      cellSizeDistribution: {
-        M: {
+      processingConfig: {
+        cellSizeDistribution: {
+          filterSettings: { minCellSize: 10800, binStep: 200 },
+          enabled: true,
```
ivababukova, line 58: @aerlaut, for dimensionality reduction is this the same format that you modified to recently in one of your changes of the UI for data integration config?

src/api/route-services/experiment.js
```diff
@@ -101,10 +101,15 @@ class ExperimentService {
...
   async updateProcessingConfig(experimentId, processingConfig) {
     const dynamodb = createDynamoDbInstance();
-    let key = { experimentId };
 
+    let key = { experimentId };
     key = convertToDynamoDbRecord(key);
-    const { updExpr, attrNames, attrValues } = configArrayToUpdateObjs('processingConfig', processingConfig);
+
+    const {
+      updExpr,
```
ivababukova, line 109: what is `updExpr`?

src/api/route-services/work-response.js
```diff
@@ -1,28 +1,17 @@
...
-
-
-    if (!res.valid) {
-      throw new Error(res.errors);
-    }
-
-    this.workResponse = workResponse;
-    this.io = io;
+    return (async () => {
+      await validateRequest(workResponse, 'WorkRequest.v1.yaml');
```
ivababukova: nice! We have to test this thoroughly though if there are any existing work responses that are not according to the schema format

ivababukova: also, shouldn't this be validated with the `WorkResponse` schema instead of the `WorkRequest`?

ivababukova: also, what happens if there is a validation error? Do we log it anywhere? We don't want the API to silently fail if there is a validation error, because it will be super hard to debug.

src/specs/models/api-response-schemas/ProcessingConfig.v1.yaml
```diff
@@ -0,0 +1,29 @@
...
+    $ref: '../processing-config-bodies/ProcessingConfigClassifier.v1.yaml'
+  numGenesVsNumUmis:
+    $ref: '../processing-config-bodies/ProcessingConfigNumGenesVsNumUMIs.v1.yaml'
+  doubletScores:
+    $ref: '../processing-config-bodies/ProcessingConfigDoubletScores.v1.yaml'
+  dimensionalityReduction:
+    $ref: '../processing-config-bodies/ProcessingConfigDimensionalityReduction.v1.yaml'
+  configureEmbedding:
+    $ref: '../processing-config-bodies/ProcessingConfigConfigureEmbedding.v1.yaml'
+required:
```
ivababukova, line 21: if all of them are required, then what will we do if the user wants to run data processing with only a subset of those filters? How are we going to support custom filters one day?

src/specs/models/ProcessingConfigBodies.v1.yaml
```diff
@@ -0,0 +1,12 @@
...
+description: Valid processing config bodies.
+anyOf:
+  - $ref: ./processing-config-bodies/ProcessingConfigCellSizeDistribution.v1.yaml
+  - $ref: ./processing-config-bodies/ProcessingConfigClassifier.v1.yaml
+  - $ref: ./processing-config-bodies/ProcessingConfigMitochondrialContent.v1.yaml
+  - $ref: ./processing-config-bodies/ProcessingConfigReadAlignment.v1.yaml
+  - $ref: ./processing-config-bodies/ProcessingConfigClassifier.v1.yaml
+  - $ref: ./processing-config-bodies/ProcessingConfigNumGenesVsNumUMIs.v1.yaml
+  - $ref: ./processing-config-bodies/ProcessingConfigDoubletScores.v1.yaml
+  - $ref: ./processing-config-bodies/ProcessingConfigConfigureEmbedding.v1.yaml
```
ivababukova, line 11: right now for the configure the embedding UI module, we only send GetEmbedding task to the worker when the user changes the embedding settings. How is this `ProcessingConfigConfigureEmbedding.v1.yaml` going to be used now?

src/specs/models/processing-config-bodies/ProcessingConfigConfigureEmbedding.v1.yaml
```diff
@@ -0,0 +1,51 @@
...
+        properties:
+          umap:
+            type: object
+            properties:
+              minimumDistance:
+                type: number
+              distanceMetric:
+                type: string
+                minLength: 1
+          tsne:
```
ivababukova, line 24: this is essentially the same config that we use for the `GetEmbedding` task -- writing it again in here means that if we want to make any changes to the schema, we will have to update it both here and in `GetEmbedding`, which is not ideal.

src/specs/models/processing-config-bodies/ProcessingConfigConfigureEmbedding.v1.yaml
```diff
@@ -0,0 +1,51 @@
...
+        type: object
+        properties:
+          louvain:
+            type: object
+            properties:
+              resolution:
+                type: number
+required:
+  - embeddingSettings
+  - clusteringSettings
```
ivababukova, line 50: in the UI, the embedding settings are treated as a different entity from clustering settings -- the user sends requests for embedding settings change when clicking the Apply button, whereas for the clustering settings, these get changed after the user moves the slider (where the actual work behind this is not implemented in the backend). Because we will never really send embedding and clustering settings in one request, I don't think that it makes sense to have them like that tied to one schema and also making them both required.

src/specs/models/processing-config-bodies/ProcessingConfigDimensionalityReduction.v1.yaml
```diff
@@ -0,0 +1,32 @@
+title: Dimensionality Reduction
```
ivababukova, line 1: this should be called Data Integration, so that it is consistent with what we have in the UI

src/specs/models/processing-config-bodies/ProcessingConfigMeta.v1.yaml
ivababukova: @aerlaut can you review this?

tests/api/routes/experiment.test.js
marcellp: No, because we are effectively testing a mock here. We mock out this function, then we test that it returns what the mock should return. What we should be testing instead is if the data complies with the spec, which is what this test suite is for.

src/specs/models/processing-config-bodies/ProcessingConfigConfigureEmbedding.v1.yaml
marcellp: I see, this is how it was designed originally, I just refactored into separate files. We can always change the schema if we need to.

src/specs/models/processing-config-bodies/ProcessingConfigDimensionalityReduction.v1.yaml
marcellp: Again, this is what it's called right now and how the default experiments are structured. If we need to change it, it should be a separate ticket probably.

src/specs/models/processing-config-bodies/ProcessingConfigConfigureEmbedding.v1.yaml
marcellp: Correct, we are not currently checking the embedding config to comply with any schema. If we were to do this, I would do it in a separate ticket as this is out of scope for this change.

src/specs/models/ProcessingConfigBodies.v1.yaml
marcellp: The only reason this was done is to ease readability from Stoplight Studio. See: 
image

src/specs/models/api-response-schemas/ProcessingConfig.v1.yaml
marcellp: This is the schema the API uses to report on the data processing step. Currently the API reports the entire tree back. 

See: https://api.scp.biomage.net/v1/experiments/5928a56c7cbff9de78974ab50765ed20/processingConfig

Again, this is just the original API schema that I refactored into a separate file.

src/api/route-services/work-response.js
marcellp: For (1), we don't actually validate based on the work response schemas right now, as we only have a limited number of them. You can see this here -- response is just any object, no requirements as to its contents.

image

marcellp: For (2), yes, that is a typo. Will fix.

marcellp: For (3), the same exact thing happens as right now when there is a validation error. The error is thrown into the handler for `/workResult`, which does not handle it, so it ends up in the main error handler of the API and logged.

src/utils/schema-validator.js
marcellp: Yes, that is how `fetch` does it, and this must be fetch-compliant.

src/api/route-services/experiment.js
marcellp: the updated expression.

src/specs/models/processing-config-bodies/ProcessingConfigMeta.v1.yaml
aerlaut: Yup, looks good. It stores data on which steps are completed, and whether all the steps are completed.

src/api/route-services/__mocks__/experiment.js
aerlaut: `dimensionalityReduction` is changed to `dataIntegration` which has the following default state : 
```
dataIntegration: {
      dataIntegration: {
        method: 'seuratv4',
        methodSettings: {
          seuratv4: {
            numGenes: 2000,
            normalisation: 'logNormalise',
          },
        },
      },
      dimensionalityReduction: {
        method: 'rpca',
        numPCs: 30,
        variationExplained: 91,
        excludeGeneCategories: [],
      },
    }
```

the key `dimensionalityReduction` is removed. This change is actually outside of this task, it should be done when persisting `dataIntegration` config to database,

src/specs/models/processing-config-bodies/ProcessingConfigDimensionalityReduction.v1.yaml
aerlaut: Yup, it should be part of persisting data integration config to DynamoDB

aerlaut:
Looks all good to me. Some changes requested to the API schema are better allocated to task on persisting `DataIntegration` in DynamoDB

marcellp:
@ivababukova approved the changes but can't approve via GitHub now.

aerlaut (approved):
Approved

PR #48, opened 2021-02-17
API functionality to create and submit the state machine to AWS/inframock

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-512

TBD

N/A

N/A

# Changes
### Code changes
Implemented the requirements set out in the ticket. There are also numerous miscellaneous modifications to other repositories which have been merged in already.

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/api/route-services/pipelines.js
+++ b/src/api/route-services/pipelines.js
@@ -174,20 +174,13 @@ class PipelinesService {
       type: 'STANDARD',
     };
 
-    console.log(params);
     try {
       const { stateMachineArn } = await this.stepFunctions.createStateMachine(params).promise();
       return { stateMachineArn };
     } catch (e) {
-      console.log('other error!');
-
-
       if (e.code !== 'StateMachineAlreadyExists') {
-        console.log('throw!@!');
         throw e;
       }
-
-      console.log('try again');
 
       const stateMachineArn = `arn:aws:states:${config.awsRegion}:${accountId}:stateMachine:${params.name}`;
 
--- a/src/api/route-services/pipelines.js
+++ b/src/api/route-services/pipelines.js
@@ -1,0 +1,203 @@
+const AWS = require('aws-sdk');
+const crypto = require('crypto');
+const fetch = require('node-fetch');
+const YAML = require('yaml');
+const jq = require('jq-web');
+const config = require('../../config');
+
+
+class PipelinesService {
+  constructor() {
+    this.stepFunctions = new AWS.StepFunctions({
+      region: config.awsRegion,
+    });
+
+    this.clusterInfo = null;
+  }
+
+  async getClusterInformation() {
+    if (!this.clusterInfo) {
+      const eks = new AWS.EKS({
+        region: config.awsRegion,
+      });
+
+      const accountId = await config.awsAccountIdPromise();
+      const info = await eks.describeCluster({ name: `biomage-${config.clusterEnv}` }).promise();
+
+      const { name, endpoint } = info.cluster;
+      const certAuthority = info.cluster.certificateAuthority.data;
+      const roleArn = `arn:aws:iam::${accountId}:role/state-machine-role-${config.clusterEnv}`;
+
+      this.clusterInfo = {
+        accountId, name, endpoint, certAuthority, roleArn,
+      };
+    }
+
+    return this.clusterInfo;
+  }
+
+  static async getPipelineImage() {
+    const response = await fetch(
+      config.pipelineInstanceConfigUrl,
+      {
+        method: 'GET',
+      },
+    );
+
+    const txt = await response.text();
+    const manifest = YAML.parseAllDocuments(txt);
+
+    return jq.json(manifest, '..|objects|.r//empty');
+  }
+
+  async generatePipeline(experimentId) {
+    const { name, endpoint, certAuthority } = await this.getClusterInformation();
+    const namespace = `worker-${config.sandboxId}`;
+    const image = await PipelinesService.getPipelineImage();
+
+    return {
+      Comment: 'N/A',
+      StartAt: 'DeleteCompletedJobs',
+      States: {
+        DeleteCompletedJobs: {
+          Type: 'Task',
+          Comment: 'Deletes all the preivous server jobs that are already completed.',
+          Resource: 'arn:aws:states:::eks:call',
+          Parameters: {
+            ClusterName: name,
+            CertificateAuthority: certAuthority,
+            Endpoint: endpoint,
+            Method: 'DELETE',
+            Path: `/apis/batch/v1/namespaces/${namespace}/jobs`,
+            QueryParameters: {
+              fieldSelector: [
+                'status.successful=1',
+              ],
+            },
+          },
+          Next: 'LaunchKubernetesJobIfNotExists',
+        },
+        LaunchKubernetesJobIfNotExists: {
+          Type: 'Task',
+          Comment: 'Attempts to create a Kubernetes Job for the pipeline server. Will swallow a 409 (already exists) error.',
+          Resource: 'arn:aws:states:::eks:call',
+          Parameters: {
+            ClusterName: name,
+            CertificateAuthority: certAuthority,
+            Endpoint: endpoint,
+            Method: 'POST',
+            Path: `/apis/batch/v1/namespaces/${namespace}/jobs`,
+            RequestBody: {
+              apiVersion: 'batch/v1',
+              kind: 'Job',
+              metadata: {
+                name: `remoter-server-${experimentId}`,
+              },
+              spec: {
+                template: {
+                  metadata: {
+                    name: `remoter-server-${experimentId}`,
+                  },
+                  spec: {
+                    containers: [
+                      {
+                        name: 'remoter-server',
+                        image,
+                      },
+                    ],
+                    restartPolicy: 'Never',
+                  },
+                },
+              },
+            },
+          },
+          Retry: [
+            {
+              ErrorEquals: ['EKS.409'],
+              IntervalSeconds: 1,
+              BackoffRate: 2.0,
+              MaxAttempts: 2,
+            },
+          ],
+          Catch: [
+            {
+              ErrorEquals: ['EKS.409'],
+              ResultPath: '$.error-info',
+              Next: 'Wait',
+            },
+          ],
+          Next: 'Wait',
+        },
+        Wait: {
+          Type: 'Wait',
+          Seconds: 5,
+          End: true,
+        },
+      },
+    };
+  }
+
+
+  async create(experimentId) {
+    const {
+      roleArn, accountId,
+    } = await this.getClusterInformation();
+
+    const { clusterEnv, sandboxId } = config;
+
+    const pipelineHash = crypto
+      .createHash('sha1')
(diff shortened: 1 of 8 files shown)


src/api/route-services/pipelines.js
```diff
@@ -0,0 +1,196 @@
...
+
+    const txt = await response.text();
+    const manifest = YAML.parseAllDocuments(txt);
+
+    return jq.json(manifest, '..|objects|.r//empty');
+  }
+
+  async generatePipeline(experimentId) {
+    const { name, endpoint, certAuthority } = await this.getClusterInformation();
+    const namespace = `worker-${config.sandboxId}`;
```
ivababukova: this should probably live in the config?

src/api/route-services/pipelines.js
```diff
@@ -0,0 +1,196 @@
...
+
+      const accountId = await config.awsAccountIdPromise();
+      const info = await eks.describeCluster({ name: `biomage-${config.clusterEnv}` }).promise();
+
+      const { name, endpoint } = info.cluster;
+      const certAuthority = info.cluster.certificateAuthority.data;
+      const roleArn = `arn:aws:iam::${accountId}:role/state-machine-role-${config.clusterEnv}`;
+
+      this.clusterInfo = {
+        accountId, name, endpoint, certAuthority, roleArn,
```
ivababukova, line 32: you only use `accountId`, `certAuthority` and `name` from `clusterInfo`, for the rest since they are not used, it is best to delete them

src/api/route-services/pipelines.js
```diff
@@ -0,0 +1,196 @@
...
+      config.pipelineInstanceConfigUrl,
+      {
+        method: 'GET',
+      },
+    );
+
+    const txt = await response.text();
+    const manifest = YAML.parseAllDocuments(txt);
+
+    return jq.json(manifest, '..|objects|.r//empty');
```
ivababukova, line 50: we need handling here on what happens if the GET request returns an error

src/api/route-services/pipelines.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  async getClusterInformation() {
+    if (!this.clusterInfo) {
+      const eks = new AWS.EKS({
+        region: config.awsRegion,
+      });
+
+      const accountId = await config.awsAccountIdPromise();
+      const info = await eks.describeCluster({ name: `biomage-${config.clusterEnv}` }).promise();
```
ivababukova, line 25: we need some error handling here

src/api/route-services/pipelines.js
```diff
@@ -0,0 +1,196 @@
...
+
+    return jq.json(manifest, '..|objects|.r//empty');
+  }
+
+  async generatePipeline(experimentId) {
+    const { name, endpoint, certAuthority } = await this.getClusterInformation();
+    const namespace = `worker-${config.sandboxId}`;
+    const image = await PipelinesService.getPipelineImage();
+
+    return {
```
ivababukova, line 57: I wonder if this is the best place to be keeping this job config?

src/api/route-services/pipelines.js
```diff
@@ -0,0 +1,196 @@
...
+      .createHash('sha1')
+      .update(`${experimentId}-${sandboxId}`)
+      .digest('hex');
+
+    const params = {
+      name: `biomage-pipeline-${pipelineHash}`,
+      roleArn,
+      definition: JSON.stringify(await this.generatePipeline(experimentId)),
+      loggingConfiguration: {
+        level: 'OFF',
```
ivababukova, line 157: Logging to where is this for? Why is it off?

src/api/route-services/pipelines.js
```diff
@@ -0,0 +1,196 @@
...
+        {
+          key: 'sandboxId',
+          value: sandboxId,
+        },
+      ],
+      type: 'STANDARD',
+    };
+
+    try {
+      const { stateMachineArn } = await this.stepFunctions.createStateMachine(params).promise();
```
ivababukova, line 177: where do you define the actual steps that the step machine will be running with?

src/config/test-config.js
```diff
@@ -11,7 +11,8 @@ module.exports = {
   clusterEnv: 'test',
   awsRegion: 'eu-west-1',
   awsAccountIdPromise: getAwsAccountId,
-  workerInstanceConfigUrl: 'https://raw.githubusercontent.com/biomage-ltd/iac/master/charts/worker-instance.yaml',
+  workerInstanceConfigUrl: 'https://raw.githubusercontent.com/biomage-ltd/iac/master/releases/production/worker.yaml',
+  pipelineInstanceConfigUrl: 'https://raw.githubusercontent.com/biomage-ltd/iac/master/releases/production/pipeline.yaml',
```
ivababukova, line 15: why do we need these with `production` in them for the test config?

src/specs/api.yaml
```diff
@@ -314,11 +314,32 @@ paths:
...
       description: Deletes a plot and table for a given experiment with the data specified.
+  '/experiments/{experimentId}/pipelines':
+    parameters:
+      - schema:
+          type: string
+        name: experimentId
+        in: path
+        required: true
+    post:
+      summary: Create a new pipeline for taks execution
```
ivababukova, line 325: typo on tasks

src/specs/api.yaml
```diff
@@ -314,11 +314,32 @@ paths:
...
+      x-eov-operation-handler: routes/pipelines
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties: {}
+      description: This path will create a new pipeline that can run a state machine with different bioinformatics tasks.
```
ivababukova, line 337: where do we supply the tasks that we want the statemachine to run?

src/api/route-services/pipelines.js
marcellp: `endpoint` and `roleArn` are also used.

src/api/route-services/pipelines.js
marcellp: The HTTP server is already rigged up with its own error handler. If there is an error getting the file, it will throw an exception and return with a 500. We can, if you want to, have alternative error handling, but since failing to fetch the config makes it impossible to launch the pipeline, we can fail here hard.

src/api/route-services/pipelines.js
ivababukova: that's true, I just saw that, all good then

src/api/route-services/pipelines.js
marcellp: Same question as before. If the cluster cannot be found, an exception will be thrown and there is no way to recover from it.

src/api/route-services/pipelines.js
marcellp: No, this will need to be refactored for local development anyways (or when a custom pipeline is run), so I thought I'd keep it here for now and find a new place for it in the next ticket.

src/api/route-services/pipelines.js
marcellp: This is for logging to CloudTrail. I don't think we need this for now, as the actual results from the steps, as well as their inputs, still get persisted in the execution.

src/api/route-services/pipelines.js
marcellp: Those are defined in `generatePipeline`. For now, the state machine is static and is hard-coded. This will have to be changed for local development, because the steps will have to be different (running Lambdas or something similar instead of Kubernetes jobs), so they will need to be refactored as part of that ticket when I am more sure about what the exact use case will be.

src/api/route-services/pipelines.js
ivababukova: as long as we make this code here to return something meaningful if there is an error from the endpoint, I think it should work. The reason why I added comment in here is because it would not be great if it this just blows up silently and then we have to go and figure out what is happening, but maybe this is part of the error handling ticket?

src/api/route-services/pipelines.js
ivababukova: ok, if that is the plan, sounds good

src/api/route-services/pipelines.js
ivababukova: yup, we don't need it for now.

src/config/test-config.js
marcellp: These are for defining things like the most up-to-date image for the given deployment, etc. The API uses that data to inform, e.g., which image to launch the pipeline server with. You can see this as part of a `ConfigMap` in the files.

These are only created for staging and production when the specific projects are deployed. We can use the production one for testing purposes so we always have some data in the files -- the actual data itself does not matter, because we don't actually call to any services, they just have to be defined.

src/specs/api.yaml
marcellp: Good catch, will fix.

src/api/route-services/pipelines.js
ivababukova: they will also need to be different because the users need to be able to exclude certain filters from the pipeline. Ok, if the plan is to update them later, sounds good.

src/specs/api.yaml
marcellp: We don't at this point, the ticket's scope does not extend to that, see https://biomage.atlassian.net/browse/BIOMAGE-512

src/api/route-services/pipelines.js
marcellp: I was thinking about that, but it is only used once, and it's directly derived from another config variable. If you feel like it's a better place for it, I'll do it though.

ivababukova: I think it is better to move it. In the future, if we decided to change the naming conventions for the namespace, we could forget about this line here and break the code, because of that I think it is better to have it all in one place.

src/api/route-services/pipelines.js
marcellp: Yes, this is part of error handling ticket.

The API always throws an error when things happen on the HTTP side. Those are always caught and an error message is raised on the client, such as the message you see when you can't connect to the API.

Errors are still caught and processed properly on the Socket.IO side as well, it's just that they are not returned back to the UI. For now, this pipeline stuff is an HTTP endpoint because it's easier to test it locally without hooking it up to the UI.

Essentially, the API never blows up under an error condition, it always keeps running, it just silently suppressed exceptions for Socket.IO events, and that is what the ticket to fix that is for.

src/api/route-services/pipelines.js
marcellp: That's right. Plan is to have some sort of map-reduce function to take the config, construct a new pipeline stepwise (with different step configurations for local and AWS development, as discussed in the morning), and put it together. This is just a placeholder for now.

src/api/route-services/pipelines.js
ivababukova: ok, sounds good!

src/api/route-services/pipelines.js
ivababukova: yup, as discussed in the other thread, sounds good

ivababukova (approved):
after the tiny things to fix, it is good

PR #57, opened 2021-03-03
Add /pipelineResponse path

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-509

https://ui-pipeline-response-path.scp-staging.biomage.net/

https://github.com/biomage-ltd/pipeline/pull/5

N/A

# Changes
### Code changes
Added new path for pipeline response parsing. Also refactored some of the SNS validation code into a separate function so we can reuse it across workers and pipelines.

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
- [x] Photo of a cute animal attached to this PR
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "object-hash": "^2.0.3",
     "sns-validator": "^0.3.4",
     "socket.io": "^2.3.0",
+    "socket.io-redis": "^6.0.1",
     "swagger-client": "^3.12.2",
     "swagger-model-validator": "^3.0.18",
     "tmp-promise": "^3.0.2",
--- a/src/api/route-services/pipeline-response.js
+++ b/src/api/route-services/pipeline-response.js
@@ -1,0 +1,41 @@
+const AWS = require('../../utils/requireAWS');
+const validateRequest = require('../../utils/schema-validator');
+
+const pipelineResponse = async (io, message) => {
+  await validateRequest(message, 'PipelineResponse.v1.yaml');
+
+  // Fail hard if there was an error.
+  const { response: { error } } = message;
+
+  if (error) {
+    throw new Error(error);
+  }
+
+  // Download output from S3.
+  const s3 = new AWS.S3();
+  const { output: { bucket, key } } = message;
+  const outputObject = await s3.getObject(
+    {
+      Bucket: bucket,
+      Key: key,
+    },
+  ).promise();
+  const output = JSON.parse(outputObject.Body.toString());
+
+  if (output.config) {
+    await validateRequest(output.config, 'ProcessingConfigBodies.v1.yaml');
+  }
+
+  // Concatenate into a proper response.
+  const response = {
+    ...message,
+    output,
+  };
+
+
+  const { input: { experimentId } } = message;
+  console.log('Sending to all clients subscribed to experiment', experimentId);
+  io.sockets.emit(`ExperimentUpdates-${experimentId}`, response);
+};
+
+module.exports = pipelineResponse;
--- a/src/api/route-services/pipeline-response.js
+++ b/src/api/route-services/pipeline-response.js
@@ -1,14 +1,16 @@
 const AWS = require('../../utils/requireAWS');
 const validateRequest = require('../../utils/schema-validator');
+const logger = require('../../utils/logging');
 
 const pipelineResponse = async (io, message) => {
   await validateRequest(message, 'PipelineResponse.v1.yaml');
 
   // Fail hard if there was an error.
-  const { response: { error } } = message;
+  const { response: { error }, input: { experimentId } } = message;
 
   if (error) {
-    throw new Error(error);
+    io.sockets.emit(`ExperimentUpdates-${experimentId}`, message);
+    return;
   }
 
   // Download output from S3.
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -50,6 +50,7 @@
     "object-hash": "^2.0.3",
     "sns-validator": "^0.3.4",
     "socket.io": "^2.3.0",
+    "socket.io-redis": "^6.0.1",
```
marcellp, line 53: This is an adapter which will allow us to scale up the API in production without having to worry about clients connected to different API instances.

See: https://socket.io/docs/v3/using-multiple-nodes/index.html

src/utils/parse-sns-message.js
```diff
@@ -0,0 +1,62 @@
...
+      throw err;
+    }
+  }
+
+  // Handle subscripton and unsubscription automatically.
+  if (config.clusterEnv !== 'development'
+    && (msg.Type === 'SubscriptionConfirmation'
+      || msg.Type === 'UnsubscribeConfirmation')) {
+    https.get(msg.SubscribeURL);
+  }
```
aerlaut, line 40: Is this part of the code is preparation for something in the future? or is it linked to another repo?

xverges: Funny, right now I was looking at the exact same part of the MR :-)

xverges (commented):
Everything looks good to me, but I do not trust myself very much :-)

aerlaut (approved):
Looks good to me

src/api/route-services/pipeline-response.js
```diff
@@ -0,0 +1,41 @@
...
+
+const pipelineResponse = async (io, message) => {
+  await validateRequest(message, 'PipelineResponse.v1.yaml');
+
+  // Fail hard if there was an error.
+  const { response: { error } } = message;
+
+  if (error) {
+    throw new Error(error);
+  }
```
ivababukova: will the UI receive this? Or will the UI just never get response for failed pipeline steps?

src/api/route-services/pipeline-response.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // Download output from S3.
+  const s3 = new AWS.S3();
```
ivababukova, line 17: don't we already use the s3 client in the API for downloading too big files from SNS? Because if we have the implementation of downloading of files in different places, it will be worth to have the implementation in one place and just reused

src/api/route-services/pipeline-response.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // Concatenate into a proper response.
+  const response = {
+    ...message,
+    output,
+  };
+
+
+  const { input: { experimentId } } = message;
+  console.log('Sending to all clients subscribed to experiment', experimentId);
```
ivababukova: use logger instead of console log

src/specs/api.yaml
```diff
@@ -101,23 +101,66 @@ paths:
...
         - work
     parameters: []
-  "/experiments/{experimentId}":
+  /pipelineResults:
+    post:
+      operationId: receivePipelineResponse
+      x-eov-operation-id: 'pipelines#response'
+      x-eov-operation-handler: routes/pipelines
+      requestBody:
+        description: The data sent by AWS SNS.
```
ivababukova: I would say `The results from the execution of a pipeline step, sent via SNS`

src/specs/models/PipelineResponse.v1.yaml
```diff
@@ -0,0 +1,43 @@
...
+      key:
+        type: string
+        description: The key within the S3 bucket where the result is located.
+  input:
+    type: object
+    properties:
+      experimentId:
+        type: string
+        description: The experiment ID for which the step was completed.
+      taskName:
```
ivababukova, line 23: it is most likely out of scope, but curious anyway: what will happen if from the UI the user assembled a pipeline with two tasks having duplicate names?

src/utils/parse-sns-message.js
ivababukova: The part of the code (lines 35 to 39) make sure that when running the API locally we don't try to get messages from the SNS topic on staging or in production, since we don't have sns locally at all.
The part of the code (lines 42 to 60) parse the message, if there was one received and get the socket io object and return them, so that the rest of the API can do whatever it needs to do with them.

src/api/route-services/pipeline-response.js
marcellp: Fixed.

src/specs/api.yaml
marcellp: Fixed!

src/api/route-services/pipeline-response.js
marcellp: Fixed so it does receive the error.

src/api/route-services/pipeline-response.js
marcellp: We sort of use a similar pattern, but that one is more involved as it involves looking through all the result objects and downloading only those which are required to be downloaded. This one is much simpler so there was not a lot of things to factor in together.

ivababukova: makes senes in that case

PR #58, opened 2021-03-04
Construct step function pipeline

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-506

N.A.

https://github.com/biomage-ltd/iac/pull/41

- Modified skeleton to fit the current data processing configuration

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/api/general-services/pipeline-manage/constructors/create-new-job-if-not-exist.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-job-if-not-exist.js
@@ -16,6 +16,7 @@ const createNewJobIfNotExist = (context, step) => {
         Payload: {
           image: pipelineImages['remoter-server'],
           name: 'pipeline-remoter-server',
+          detached: true,
         },
       },
     };
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -25,6 +25,7 @@ const createNewStep = (context, step, args) => {
           image: pipelineImages['remoter-client'],
           name: 'pipeline-remoter-client',
           task,
+          detached: false,
         },
       },
     };
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -36,7 +36,7 @@ const createNewStep = (context, step, args) => {
     ...step,
     Type: 'Task',
     Comment: 'Attempts to create a Kubernetes Job for the pipeline server. Will swallow a 409 (already exists) error.',
-    Resource: 'arn:aws:states:::eks:call',
+    Resource: 'arn:aws:states:::eks:runJob.sync',
     Parameters: {
       ClusterName: clusterInfo.name,
       CertificateAuthority: clusterInfo.certAuthority,
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -11,7 +11,7 @@ const createNewStep = (context, step, args) => {
   const task = JSON.stringify({
     experimentId,
     taskName,
-    config: processingConfig[taskName],
+    config: processingConfig[taskName] || {},
   });
 
   if (config.clusterEnv === 'development') {
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -1,0 +1,90 @@
+const config = require('../../../../config');
+
+const createNewStep = (context, step, args) => {
+  const {
+    processingConfig, clusterInfo, experimentId, pipelineImages, accountId,
+  } = context;
+
+  const { taskName } = args;
+
+
+  const task = JSON.stringify({
+    experimentId,
+    taskName,
+    config: processingConfig[taskName],
+  });
+
+  if (config.clusterEnv === 'development') {
+    return {
+      ...step,
+      Type: 'Task',
+      Resource: 'arn:aws:states:::lambda:invoke',
+      Parameters: {
+        FunctionName: `arn:aws:lambda:eu-west-1:${accountId}:function:local-container-launcher`,
+        Payload: {
+          image: pipelineImages['remoter-client'],
+          name: 'pipeline-remoter-client',
+          task,
+        },
+      },
+    };
+  }
+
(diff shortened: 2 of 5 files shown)


src/api/general-services/pipeline-manage/constructors/create-new-job-if-not-exist.js
```diff
@@ -16,6 +16,7 @@ const createNewJobIfNotExist = (context, step) => {
         Payload: {
           image: pipelineImages['remoter-server'],
           name: 'pipeline-remoter-server',
+          detached: true,
```
ivababukova, line 19: what is this for?

aerlaut: detached is used in local environment by the [lambda function which sets up the containers to run the tasks](https://github.com/biomage-ltd/pipeline/blob/bb7286f3ce2864b5466a6c51c139135538a92324/local-runner/cf-local-container-launcher.yaml#L32). It acts as a flag to start in detached mode or not.

`remoter-server` can be started as daemon, hence it's set to true. `remoter-client` has to be started in non-detached mode so that it runs sequentially as defined in the step functions definition.

ivababukova: would the user be the one that decides if the container should be running in the background (i.e. in detached mode)? In which cases in particular will we want to run in detached mode?

src/api/general-services/pipeline-manage/constructors/create-new-step.js
```diff
@@ -0,0 +1,91 @@
...
+        metadata: {
+          name: `remoter-client-${experimentId}`,
+        },
+        spec: {
+          template: {
+            metadata: {
+              name: `remoter-client-${experimentId}`,
+            },
+            spec: {
+              containers: [
```
ivababukova: doesn't a lot of this overlap with the object that `createNewJobIfNotExist` returns? Have you checked if it makes sense to extract the common logic in one place? Or is it going to lead to leaky abstraction?

src/api/general-services/pipeline-manage/constructors/create-new-step.js
```diff
@@ -0,0 +1,91 @@
+const config = require('../../../../config');
+
+const createNewStep = (context, step, args) => {
```
ivababukova, line 3: I see that this function is not used anywhere at the moment -- are we going to start using it after the UI bits for QC are implemented too?

src/api/general-services/pipeline-manage/index.js
```diff
@@ -41,17 +44,21 @@ const getClusterInfo = async () => {
...
-      return f(context, step);
+      return f(context, step, args);
     }
     case 'create-new-job-if-not-exist': {
       const f = require('./constructors/create-new-job-if-not-exist');
-      return f(context, step);
+      return f(context, step, args);
+    }
+    case 'create-new-step': {
+      const f = require('./constructors/create-new-step');
```
ivababukova: I see that the `create-new-step` function is used in here actually -- better to import `createNewStep` and call `createNewStep(context, step, args)`, that will make the code easier to read. I would recommend doing this for the other two cases too

ivababukova:
What did you do to test that this works?

src/api/general-services/pipeline-manage/constructors/create-new-job-if-not-exist.js
aerlaut: No, this setting is set internally, not by the user. As I understand it, it's to enforce sequential creation and execution of the `remoter-client` containers, so the steps are executed in order.

src/api/general-services/pipeline-manage/constructors/create-new-step.js
aerlaut: It's used in `pipeline-manage/index.js`,

src/api/general-services/pipeline-manage/constructors/create-new-job-if-not-exist.js
ivababukova: I see, thanks for explaining!

src/api/general-services/pipeline-manage/constructors/create-new-step.js
marcellp: Actually, great point Iva.

The leaky abstraction point is not going to be an issue imho. This is becausethe `createNewJobIfNotExist` function's only purpose is to *create* a server. It does not monitor its lifecycle as it's meant to be run long-term.

For this constructor (`create-new-step`), there is a proper way to run a Job and wait for it to finish -- this is documented [here](https://docs.aws.amazon.com/step-functions/latest/dg/connect-eks.html). Instead of using `arn:aws:states::eks:call` and just create a Job like we do with the server (which will run throughout the pipeline), the steps themselves should be executed using `arn:aws:states:::eks:runJob.sync`.

This makes it possible for the state machine to wait for the Job to exit successfully or with an error and stop the machine if there has been an error. It also will save the logs that the client outputs in the Step Function which we can review or save to CloudWatch later if we want to.

We should use the special-purpose `arn:aws:states:::eks:runJob.sync` for `create-new-step.js` accordingly. Most of the code is basically the same so it shouldn't require too much changing.

marcellp (changes requested):

src/api/general-services/pipeline-manage/constructors/create-new-step.js
aerlaut: Changed in commit 344f4bf

aerlaut:
> What did you do to test that this works?

Tested locally. Added test to check for the result of merging the config from DynamoDB.

aerlaut:
Btw, with the current implementation, even though users have just finished 1 step, all the steps will be run. Is this expected?

ivababukova (approved):

ivababukova:
> Btw, with the current implementation, even though users have just finished 1 step, all the steps will be run. Is this expected?

Depends on whether the implementation for support of custom pipeline creation should be done as part of this ticket or not. Judging from this ticket: https://biomage.atlassian.net/browse/BIOMAGE-506 in the goals it says that it should. But since this PR is ready to go, it might be a better idea to merge it as it is and make the pipeline be custom as part of a separate ticket

aerlaut:
Creation of custom pipeline will be moved to a separate ticket

PR #79, opened 2021-03-26
Cellscope to support multi-sample data sets

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-521

https://ui-rbbtb4equafqp88h43eyfr7o8h.scp-staging.biomage.net/

https://github.com/biomage-ltd/pipeline/pull/19

N/A

# Changes
### Code changes
Added some logic to handle multisample data sets, including properly replacing processing config changes and creating full configs per sample.

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -19,16 +19,6 @@ const createNewStep = (context, step, args) => {
     server: remoterServer,
   });
 
-  // If enabled is defined and is false.
-  if (config.enabled === false) {
-    return {
-      ...step,
-      Type: 'Pass',
-      ResultPath: null,
-
-    };
-  }
-
   const stepHash = crypto.randomBytes(5).toString('hex');
 
   if (config.clusterEnv === 'development') {
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -36,7 +36,6 @@ const createNewStep = (context, step, args) => {
           name: 'pipeline-remoter-client',
           task,
           ...perSample && { 'sampleUuid.$': '$.sampleUuid' },
-          ...!perSample && { sampleUuid: 'all' },
           detached: false,
         },
       },
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -19,6 +19,16 @@ const createNewStep = (context, step, args) => {
     server: remoterServer,
   });
 
+  // If enabled is defined and is false.
+  if (config.enabled === false) {
+    return {
+      ...step,
+      Type: 'Pass',
+      ResultPath: null,
+
+    };
+  }
+
   const stepHash = crypto.randomBytes(5).toString('hex');
 
   if (config.clusterEnv === 'development') {
--- a/src/api/general-services/pipeline-manage/constructors/create-new-step.js
+++ b/src/api/general-services/pipeline-manage/constructors/create-new-step.js
@@ -98,11 +98,7 @@ const createNewStep = (context, step, args) => {
                     task,
                   ],
                   env: [
-                    {
-                      name: 'SAMPLE_ID',
-                      ...perSample && { 'value.$': '$.sampleUuid' },
-                      ...!perSample && { value: 'all' },
-                    },
+                    ...perSample ? [{ name: 'SAMPLE_ID', 'value.$': '$.sampleUuid' }] : [],
                   ],
                 },
               ],
--- a/src/api/general-services/pipeline-manage/index.js
+++ b/src/api/general-services/pipeline-manage/index.js
@@ -226,8 +226,7 @@ const createPipeline = async (experimentId, processingConfigUpdates) => {
   const roleArn = `arn:aws:iam::${accountId}:role/state-machine-role-${config.clusterEnv}`;
 
   logger.log(`Fetching processing settings for ${experimentId}`);
-  const processingRes = await experimentService.getProcessingConfig(experimentId);
-  const { processingConfig } = processingRes;
+  const { processingConfig } = await experimentService.getProcessingConfig(experimentId);
 
   const samplesRes = await samplesService.getSampleIds(experimentId);
   const { samples } = samplesRes;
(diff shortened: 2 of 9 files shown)


src/api/general-services/pipeline-manage/index.js
```diff
@@ -244,13 +243,35 @@ const createPipeline = async (experimentId, processingConfigUpdates) => {
...
+      const result = {};
+
+      samples.ids.forEach((sample) => {
+        result[sample] = _.merge(rawConfig, o[sample]);
+      });
+
+      return result;
+    }
+  });
+
```
marcellp: This takes a processing config like:

```
{
      processingConfig: {
        cellSizeDistribution: {
          filterSettings: { binStep: 200, minCellSize: 420 }, enabled: true,
        },
      },
    }
```

and turns it into something like:

```
{
  processingConfig: {
    cellSizeDistribution: {
      'sample-control': {
        filterSettings: { binStep: 200, minCellSize: 420 }, enabled: true,
      },
      'sample-treated': {
        filterSettings: { binStep: 200, minCellSize: 420 }, enabled: true,
      },
    },
  },
}
```

Where each object is a fully valid config on its own. It will also take any partial config that may be submitted specifically for a sample (sample-specific configs need not contain all required fields) and merges them into the "root" config file to create a fully valid processing config for each sample.

src/api/route-services/pipeline-response.js
```diff
@@ -4,16 +4,17 @@ const logger = require('../../utils/logging');
...
-const experimentService = new ExperimentService();
 const plotsTableService = new PlotsTablesService();
+const experimentService = new ExperimentService();
 
 const pipelineResponse = async (io, message) => {
   await validateRequest(message, 'PipelineResponse.v1.yaml');
 
   // Fail hard if there was an error.
-  const { response: { error }, input: { experimentId, taskName } } = message;
+  const { response: { error }, input: { experimentId, taskName, sampleUuid } } = message;
```
marcellp, line 17: `sampleUuid` is added as a task input so we can associate a result with a particular sample uuid (previously the result had no reference to the sample it was run on)

src/api/route-services/pipeline-response.js
```diff
@@ -53,7 +54,26 @@ const pipelineResponse = async (io, message) => {
...
+    ]);
+  } else {
+    await experimentService.updateProcessingConfig(experimentId, [
+      {
+        name: taskName,
+        body: output.config,
+      },
+    ]);
+  }
+
```
marcellp, line 76: If the task is not per-sample, just overwrite the current config as we did previously. If it is, create or update the sample-specific config with the result.

src/api/general-services/pipeline-manage/constructors/create-new-step.js
```diff
@@ -99,11 +98,7 @@ const createNewStep = (context, step, args) => {
                     task,
                   ],
                   env: [
-                    {
-                      name: 'SAMPLE_ID',
-                      ...perSample && { 'value.$': '$.sampleUuid' },
-                      ...!perSample && { value: 'all' },
-                    },
+                    ...perSample ? [{ name: 'SAMPLE_ID', 'value.$': '$.sampleUuid' }] : [],
```
marcellp, line 101: The default `SAMPLE_ID` if none is specified (i.e. the task is not sample-specific) should be no env variable at all (pipeline will convert that into an empty string, which is what R expects).

src/api/general-services/pipeline-manage/constructors/create-new-step.js
```diff
@@ -36,7 +36,6 @@ const createNewStep = (context, step, args) => {
           name: 'pipeline-remoter-client',
           task,
           ...perSample && { 'sampleUuid.$': '$.sampleUuid' },
-          ...!perSample && { sampleUuid: 'all' },
```
marcellp, line 39: The default `SAMPLE_ID` if none is specified (i.e. the task is not sample-specific) should be no env variable at all (pipeline will convert that into an empty string, which is what R expects).

src/api/general-services/pipeline-manage/index.js
```diff
@@ -244,13 +243,35 @@ const createPipeline = async (experimentId, processingConfigUpdates) => {
     });
   }
 
+
+  // This is the processing configuration merged for multiple samples where
+  // appropriate.
+  // eslint-disable-next-line consistent-return
+  const mergedProcessingConfig = _.cloneDeepWith(processingConfig, (o) => {
+    if (_.isObject(o) && !o.dataIntegration && o.enabled) {
```
cosa65, line 250: Should we check for !o.configureEmbedding?

src/api/general-services/pipeline-manage/index.js
cosa65: Would it make sense to handle enabled/disabled on the api side? By perhaps generating the skeleton for the steps according to the enabled/disabled steps?

src/api/general-services/pipeline-manage/index.js
marcellp: It is already encompassed by `enable` not being present on that one.

marcellp:
> Would it make sense to handle enabled/disabled on the api side? By perhaps generating the skeleton for the steps according to the enabled/disabled steps?

Yes, it probably would actually, especially given right now those filters take forever to run even if they are empty. I'll try to do that on Monday.

marcellp:
It seems there is a bug, the API does not update with completed tasks.

image
image

marcellp:
@cosa65 -- Disabled states no longer get executed.

src/api/general-services/pipeline-status.js
```diff
@@ -117,16 +116,29 @@ const getPipelineStatus = async (experimentId) => {
...
+      const history = await stepFunctions.getExecutionHistory({
+        executionArn,
+        includeExecutionData: false,
+        nextToken,
+      }).promise();
+
+      events = [...events, ...history.events];
+      nextToken = history.nextToken;
+    } while (nextToken);
+    /* eslint-enable no-await-in-loop */
```
marcellp, line 138: Fix pagination error that caused longer executions to not show up as complete.

src/api/general-services/pipeline-manage/constructors/create-new-step.js
```diff
@@ -19,6 +19,16 @@ const createNewStep = (context, step, args) => {
...
+  // If enabled is defined and is false.
+  if (config.enabled === false) {
+    return {
+      ...step,
+      Type: 'Pass',
+      ResultPath: null,
+
+    };
+  }
+
```
marcellp: Replace with a `Pass` state if the step is not enabled to speed up execution in those cases.

xverges: In this case, it won't be reported as completed. Only `Tasks` are. If we change that, we will need also to filter out for the UI `EndOfMap` and `EndOfPipeline`

xverges (approved):
lgtm, but make sure you are aware of my comment on `Pass` states.

cosa65 (approved):

src/api/general-services/pipeline-manage/constructors/create-new-step.js
marcellp: hmmm, okay, I think I will just delete this change then for now.

PR #105, opened 2021-05-08
Get data upload status

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-713

UI - https://github.com/biomage-ltd/ui/pull/245
API - https://github.com/biomage-ltd/api/pull/105
IAC - https://github.com/biomage-ltd/iac/pull/110

- Created new actions to save project and samples data to DynamoDB
- Staging in UI repo

# Changes
### Code changes

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/specs/api.yaml
```diff
@@ -5,7 +5,7 @@ info:
   description: The main Node.js pipeline API.
   license:
     name: MIT
-    url: "https://github.com/biomage-ltd/api/blob/master/LICENSE"
```
xverges, line 8: @aerlaut I thought that you and some other team member had conflicting setting to edit these yaml files, but I saw that it is you against you :-) https://github.com/biomage-ltd/api/commit/91eee5841b8967338e96d1ad21db240b3a9ec2a1#diff-9590aca4a81dce3ffd13c1d719b7c87a7e4c3e6a82b8fbd930e3ab1da22b4aba

Can you try to find out why this recurring `"` -> `'` -> `"` -> ... happens? Maybe we need to add some setting to the workspace?
https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml <- this is the extension that is recommended in our onboarding docs, and has a single quote setting

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Node.js API for the single-cell pipeline.",
   "main": "index.js",
   "scripts": {
-    "test": "jest --silent",
+    "test": "jest",
     "test:loud": "jest",
```
xverges: Do you think "loud" should be the default? If that is the case, should there be a `test:quiet` instead of the current `test:loud`?

src/api/route-services/__mocks__/projects.js
```diff
@@ -0,0 +1,24 @@
...
+  ) {
+    throw new Error('Invalid body');
+  }
+
+  return Promise.resolve({
+    name: 'Test project',
+    description: '',
+    createdDate: '',
+    lastModified: '',
+    uuid: 'project-1',
```
xverges: Shouldn't this be the `projectUuid` that we got as a parameter?

src/specs/api.yaml
```diff
@@ -405,16 +405,109 @@ paths:
...
+        '404':
+          description: Invalid projectUuid.
+      description: Create or update a project
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: ./models/api-body-schemas/Project.v1.yaml
+        description: The new configuration to update the old one by.
+  '/projects/{projectUuid}/samples':
```
xverges: Aren't 404 status also needed for operations on sample resources?

src/specs/models/api-body-schemas/Project.v1.yaml
```diff
@@ -0,0 +1,28 @@
+type: object
+title: ''
+description: Schema for a sample object
```
xverges: Wrong description

tests/api/routes/samples.test.js
```diff
@@ -24,18 +21,69 @@ describe('tests for samples route', () => {
...
+      .end((err) => {
+        if (err) {
+          return done(err);
+        }
+        return done();
+      });
+  });
+
+
+  it('Updating samples send error 500 if body is invalid', async (done) => {
```
xverges: `500` indicates that the server is broken. When the request is not well constructed, we need to let the client know. A `400` would be more appropriate https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses 

In fact, this is being done for projects

xverges (changes requested):
Do not return status 500 for invalid input

src/specs/api.yaml
```diff
@@ -405,16 +405,109 @@ paths:
...
+                  type: string
+                samples:
+                  $ref: ./models/api-body-schemas/Samples.v1.yaml
+    get:
+      summary: 'Get samples under a project '
+      operationId: getProjectSamples
+      x-eov-operation-id: samples#get
+      x-eov-operation-handler: routes/samples
+      responses:
+        '200':
```
ivababukova, line 497: the response can also be 404 if the sample is not found

src/specs/models/api-body-schemas/Project.v1.yaml
```diff
@@ -0,0 +1,28 @@
...
+properties:
+  name:
+    type: string
+  description:
+    type: string
+  createdDate:
+    type: string
+  lastModified:
+    type: string
+  uuid:
```
ivababukova, line 13: I would call this `projectUuid` to make it less ambigious

tests/api/route-services/projects.test.js
```diff
@@ -0,0 +1,57 @@
...
+    const getItemSpy = jest.fn((x) => x);
+    AWSMock.setSDKInstance(AWS);
+    AWSMock.mock('DynamoDB', 'updateItem', (params, callback) => {
+      getItemSpy(params);
+      callback(null, dynamodbData);
+    });
+    return getItemSpy;
+  };
+
+  it('udpateProject work', async (done) => {
```
ivababukova: typos here

tests/api/route-services/samples.test.js
```diff
@@ -54,15 +83,46 @@ describe('tests for the samples service', () => {
...
           TableName: 'samples-test',
-          Key: { experimentId: { S: '12345' } },
+          Key: { experimentId: { S: 'project-1' } },
           ProjectionExpression: 'samples.ids',
         });
       })
       .then(() => done());
   });
+
+  it('udpateSamples work', async (done) => {
```
ivababukova: typo in `udpateSamples ` too

ivababukova: apart from unit test for whether the update works, we should test for the behaviour for any corner cases that are possible, for example: What if we try to update non-existent sample? What if we pass empty samples entry in the update: is it going to delete all existent samples? What if dynamoDB was down and the actual update didn't happen?

tests/api/routes/projects.test.js
```diff
@@ -0,0 +1,76 @@
...
+    request(app)
+      .put('/v1/projects/someId')
+      .send(payload)
+      .expect(200)
+      .end((err) => {
+        if (err) {
+          console.log(err);
+          return done(err);
+        }
+        // there is no point testing for the values of the response body
```
ivababukova: is it worth testing that if there is something wrong, the schema will catch it though? Like, write a test that passes misformatted input and expects the schema validation to fail

src/api/route-services/samples.js
```diff
@@ -9,19 +9,18 @@ class SamplesService {
...
-    });
-
+  async getSamples(projectUuid) {
     const params = {
       TableName: this.tableName,
-      Key: key,
+      KeyConditionExpression: 'projectUuid = :projectUuid',
+      ExpressionAttributeValues: {
+        ':projectUuid': { S: projectUuid },
+      },
```
ivababukova: It might make more sense to reuse the dynamodb marshall and unmarshall functions for this instead of manually formatting the input yourself

src/api/route-services/samples.js
```diff
@@ -52,6 +51,31 @@ class SamplesService {
...
+    const marshalledData = convertToDynamoDbRecord({
+      ':samples': body.samples,
+      ':projectUuid': projectUuid,
+    });
+
+    // Update samples
+    const params = {
+      TableName: this.tableName,
+      Key: {
+        experimentId: { S: body.experimentId },
```
ivababukova: It might make more sense to reuse the dynamodb marshall and unmarshall functions for this instead of manually formatting the input yourself

package.json
aerlaut: Aah I see. I didn't see the `test:loud` before. I'll return this.

src/api/route-services/__mocks__/projects.js
aerlaut: No, the `projectUuid` inside a `project` is only `uuid`, so that you wouldn't be writing `project.projectUuid`.

tests/api/routes/samples.test.js
aerlaut: Ok, 400 seems a more fitting error. I'll change it

src/specs/models/api-body-schemas/Project.v1.yaml
aerlaut: As things currently are, `uuid` for an object (e.g. project or samples) refer to itself, so we wouldn't be doing `project.projectUuid` or `samples.sampleUuid`. This is how things are currently done in the API and UI. If we're going to make this change, then it's better change both in the UI and API. What do you think?

src/specs/api.yaml
aerlaut: Ok, will return this

tests/api/routes/projects.test.js
aerlaut: I think the tests for routes only checks for the existence and correct HTTP response of the routes, but does not check for what happens inside it, which is handed with tests in `route-services`.

src/api/route-services/samples.js
aerlaut: Yea, you're right, it's a risk to do it manually. I'll change it to use AWS's marshall.

src/specs/api.yaml
aerlaut: After checking`api.code-workspace`, there doesn't seem to be a rule enforcing single or double quote. I'm putting. We're using single quotes for JS files, so I'll put that rule in for YAML too.

src/specs/api.yaml
aerlaut: Added in new commit

src/specs/models/api-body-schemas/Project.v1.yaml
ivababukova: let's leave it as uuid then

src/api/route-services/projects.js
```diff
@@ -0,0 +1,42 @@
...
+} = require('../../utils/dynamoDb');
+
+const { OK, NotFoundError } = require('../../utils/responses');
+
+class ProjectsService {
+  constructor() {
+    this.tableName = `projects-${config.clusterEnv}`;
+  }
+
+  async updateProject(projectUuid, project) {
```
ivababukova, line 40: what is `project` equal to? Is it project name?

src/api/route-services/samples.js
```diff
@@ -9,37 +11,39 @@ class SamplesService {
...
+      KeyConditionExpression: 'projectUuid = :projectUuid',
+      ExpressionAttributeValues: marshalledData,
       ProjectionExpression: 'samples',
     };
     const dynamodb = createDynamoDbInstance();
 
-    const response = await dynamodb.getItem(params).promise();
+    const response = await dynamodb.query(params).promise();
 
     if (response.Item) {
```
ivababukova, line 32: Is there a chance that `response` will be null/none? Because if that happens, line 30 will crash and we should handle that case.

ivababukova: the same situation on line 51 in this file, but I can't add a comment there because it hasn't been changed

src/api/route-services/samples.js
```diff
@@ -9,37 +11,39 @@ class SamplesService {
...
       return prettyResponse;
     }
 
-    throw Error('Sample not found');
+    throw new NotFoundError('Samples not found');
   }
 
-  async getSampleIds(experimentId) {
-    const key = convertToDynamoDbRecord({
+  async getByExperimentId(experimentId) {
```
ivababukova: if you are returning samples for an experiment, it is worth renaming that function to something like `getSampleIdsForExperiment` to make it more explicit

src/api/route-services/samples.js
```diff
@@ -50,7 +54,36 @@ class SamplesService {
...
+    const marshalledKey = convertToDynamoDbRecord({
+      experimentId: body.experimentId,
+    });
+
+    const marshalledData = convertToDynamoDbRecord({
+      ':samples': body.samples,
+      ':projectUuid': projectUuid,
+    });
+
+    // Update samples
```
ivababukova, line 76: do we have samples copied across different tables? Because if we do, then we will need to update all places where we store samples

src/api/route-services/projects.js
ivababukova, line 40: do we have project data copied across different tables? Because if we do, then we will need to update all places where we store projects, so that tables don't get out of sync

src/utils/responses/index.js
```diff
@@ -0,0 +1,7 @@
+const NotFoundError = require('./NotFoundError');
```
ivababukova, line 1: what if there is a 5xx error?

tests/api/route-services/projects.test.js
```diff
@@ -0,0 +1,79 @@
...
+      Attributes: AWS.DynamoDB.Converter.marshall({ projects: jsData }),
+    };
+    const getItemSpy = jest.fn((x) => x);
+    AWSMock.setSDKInstance(AWS);
+    AWSMock.mock('DynamoDB', 'updateItem', (params, callback) => {
+      getItemSpy(params);
+      callback(null, dynamodbData);
+    });
+    return getItemSpy;
+  };
```
ivababukova: we have this piece of code everywhere where we need to mock DynamoDB -- it might be worth moving it to a single place and reusing it in this and other test files.

ivababukova: you can check in jest documentation how you can set it up. Maybe this: https://jestjs.io/docs/mock-functions#mocking-modules and this: https://jestjs.io/docs/manual-mocks and this: https://stackoverflow.com/questions/49685265/jest-mock-user-module-in-all-test-files articles can be of help

src/api/route-services/projects.js
aerlaut: a `project` object that contains `uuid`, `samples`, etc.

aerlaut: No, `project` data is only in `projects` table

src/utils/responses/index.js
aerlaut: Currently it's not custom handled - the error just bubble up to the middleware. According to the error handling guide https://expressjs.com/en/guide/error-handling.html, errors which are 4xx and 5xx are bubbled accordingly, while errors outside these ranges are turned into error 500.

src/api/route-services/samples.js
aerlaut: No, `project` data is only in `projects` table

src/api/route-services/samples.js
aerlaut: As suggested, changed to `getSamplesByExperimentId`

src/utils/responses/index.js
aerlaut: This `NotFoundError` is a custom error to handle specific conditions, so I'm not trying to catch all errors.

tests/api/route-services/projects.test.js
aerlaut: Refactored into a separate module `test-utils/mockAWSServices.js`. Provides general test modules, custom behavior still has to be written manually.

src/api/route-services/samples.js
aerlaut: Hmm... I don't think so. It should return an error `503` if there's no connection. This should be thrown at await, so it shouldn't continue to line 30. The error would bubble up through the middleware.

tests/api/route-services/projects.test.js
ivababukova: sounds great!

src/utils/responses/index.js
ivababukova: ah cool, makes sense!

src/api/route-services/samples.js
ivababukova: in this case, we should be ok.

ivababukova (approved):
looks good :) It's merging has to be coordinated with the merging of the UI though: https://github.com/biomage-ltd/ui/pull/245/files

tests/test-utils/mockAWSServices.js
```diff
@@ -0,0 +1,83 @@
+const AWSMock = require('aws-sdk-mock');
```
xverges, line 1: Love that you extracted this!

tests/test-utils/mockAWSServices.js
```diff
@@ -0,0 +1,83 @@
...
+const AWS = require('../../src/utils/requireAWS');
+
+const mockDynamoGetItem = (payload = {}, error = null) => {
+  const dynamodbData = {
+    Item: AWS.DynamoDB.Converter.marshall(payload),
+  };
+  const getItemSpy = jest.fn((x) => x);
```
xverges: Minor issue: `getItemSpy` name has been copied to the rest of non-getItem methods. Maybe you can rename all to `spy` or similar?

xverges (approved):
See minor concern about variable names in mock methods

tests/test-utils/mockAWSServices.js
aerlaut: changed to `fnSpy`

PR #112, opened 2021-05-14
using iac/master as chart reference when deploying develop branches

kafkasl:
Api is the only repo whose charts are in iac thus when trying to fetch them from iac develop branch (during staging, where STAGING_CHART_REF is `develop`) it fails because that repo has no develop branch.

### DIFF
--- a/.github/workflows/ci-develop.yaml
+++ b/.github/workflows/ci-develop.yaml
@@ -145,8 +145,14 @@ jobs:
           fi
 
           if [ "${{ matrix.environment }}" = "staging" ]; then
+            # api's chart is in IAC repo (which does not have develop branch) so if we are
+            # releasing a develop branch, it should use the master IAC repo branch
+            if [ "STAGING_CHART_REF" = "develop" ]; then
+              CHART_REF="master"
+            else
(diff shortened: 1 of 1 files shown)


ivababukova (commented):
In my opinion, the cleanest thing to do with this is to move everything to develop branch. Happy to hear other people's views, let's first discuss this and depending on the decision we can merge or not merge this change.

xverges (approved):

xverges:
@ivababukova If I understand correctly, this fix bypasses having the api charts in iac rather than in api itself. And maybe changing this is something that should be done no matter the `develop`/`master` changes

ivababukova:
My understanding is slightly different: it is that this code changes the value of `CHART_REF` to master, so that the charts get saved to the master branch of iac instead of develop, because the charts of the api live in iac and iac uses the master branch, not develop.

ivababukova:
but I think you're right -- we can merge it, then if we decided to change everything to develop, we can always fix this code here

ivababukova (approved):

kafkasl:
This was actually not the issue. It is fixed in `biomage-utils`.

PR #121, opened 2021-05-21
Recover settings accidentally removed

xverges:


ivababukova (approved):

aerlaut:
Shouldn't the settings be under 'api.code-workspace' instead of 'settings.json'?

xverges:
Agi, they were lost in a failed attempt to move them from api.code-workspace' to 'settings.json', so that multiple projects could be opened in the same window and still have proper settings.

PR #169, opened 2021-07-02
fixes for local dev

kafkasl:
mocking backend success response when runnning locally and having pulled an experiment

### DIFF
--- a/src/api/general-services/pipeline-status.js
+++ b/src/api/general-services/pipeline-status.js
@@ -11,9 +11,6 @@ const privateSteps = [
   'DeleteCompletedGem2SWorker', 'LaunchNewGem2SWorker',
 ];
 
-// error code returned when an state machine execution arn does not exist (happens when
-// pulling or moving experiments across environments)
-const EXECUTION_DOES_NOT_EXIST = 'ExecutionDoesNotExist';
 
 const notCreatedStatus = {
   startDate: null,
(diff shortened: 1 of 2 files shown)


src/api/general-services/pipeline-status.js
```diff
@@ -11,6 +11,48 @@ const privateSteps = [
   'DeleteCompletedGem2SWorker', 'LaunchNewGem2SWorker',
 ];
 
+// error code returned when an state machine execution arn does not exist (happens when
+// pulling or moving experiments across environments)
+const EXECUTION_DOES_NOT_EXIST = 'ExecutionDoesNotExist';
```
cosa65: perhas we could call it `MOCK_EXECUTION_DOES_NOT_EXIST` or `DEV_EXECUTION_DOES_NOT_EXIST` to make it clearer when it is (and should be) used. And do you think it would fit in constants.js?

cosa65 (approved):
Looks good, left one comment that you can feel free to ignore if you disagree.
Tested locally.

src/api/general-services/pipeline-status.js
kafkasl: I moved it to the constants file. However, I did not rename it because it is a "real" error. We rely on it when mocking backend status but the error itself is not a mock and in some cases it could just happen normally (if somehow the state machines description is lost or the ARN changed) and naming it "mock" could create confusion.

PR #188, opened 2021-07-23
[BIOMAGE-1137] - Order samples as defined in project

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1137

In UI repo

UI - https://github.com/biomage-ltd/ui/pull/408
API - https://github.com/biomage-ltd/api/pull/188

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


.github/workflows/ci.yaml
```diff
@@ -249,7 +249,7 @@ jobs:
 
       - name: Change name of deployment file for staging deployment
         if:
-          github.event_name == 'push' && matrix.environment == 'staging'
+          (github.event_name == 'pull_request' || github.event_name == 'push') && matrix.environment == 'staging'
```
cosa65: what is this change? Doesn't seem related to your ticket.

src/api/route-services/gem2s.js
```diff
@@ -49,7 +49,12 @@ class Gem2sService {
   static async generateGem2sTaskParams(experimentId) {
     const experiment = await (new ExperimentService()).getExperimentData(experimentId);
     const { samples } = await (new SamplesService()).getSamplesByExperimentId(experimentId);
-    const { metadataKeys } = await (new ProjectService()).getProject(experiment.projectId);
+    const {
+      metadataKeys,
+      samples: sampleIdsInOrder,
+    } = await new ProjectService().getProject(experiment.projectId);
+
+    console.log(sampleIdsInOrder);
```
cosa65: Remove this consolelog

src/api/route-services/gem2s.js
```diff
@@ -49,7 +49,12 @@ class Gem2sService {
   static async generateGem2sTaskParams(experimentId) {
     const experiment = await (new ExperimentService()).getExperimentData(experimentId);
     const { samples } = await (new SamplesService()).getSamplesByExperimentId(experimentId);
-    const { metadataKeys } = await (new ProjectService()).getProject(experiment.projectId);
+    const {
```
cosa65, line 52: it re runs gem23s if we re order the samples and then click launch analysis, I think we could avoid this relatively easily.

In the api we would need to change the way we create the hash to check if we should re run or not the pipeline, one option could be by ordering the samples before generating the hash. (just ordering for the hash so that it isn’t order-sensitive, but not ordering the actual samples array that is going to be stored)

aerlaut: Yup, this would fix the issue. The first time the project is launched after this update, it would re run GEM2S. However, on subsequent runs, it wouldn't.

src/api/route-services/gem2s.js
aerlaut: Removed

src/api/route-services/gem2s.js
aerlaut: It seems that the problem is bigger than this. Suggests moving this to another ticket.

Creating a separate variable to hold sample-order-invariant params as the source for hashing can bypass rerunning. However, the source of sample order in the UI is `cellSets`, and not rerunning GEM2S does not rearrange the order in cellSets. This necessitates other approaches such as: 

*Getting the list of experiment from `projects` in data-processing*
We don't yet have a function to get a single project, given an experimentId. Aside from that, loading projects depend on the currently logged in userId. This will be an issue down the line if we decided to support sharing of experiments.

*Storing the list of samples under an experiment object*
`experimentSettings.info` is updated everytime an experiment is launched. This object is used to store experiment configuration. We can easily pass the samples list to this object to be used later on down the line.

In data processing, sampleIds (`sampleKeys`) are derived from cellSets object. This [object actually stores the order the samples are inserted](https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/). Hence, we need to rewrite `sampleKeys` to get the data from `experimentSettings.info`.

This is all well for experiments that are launched from Data Management. However, if we refresh the page, `experimentSettings.info.samples` is missing. Upon refresh `experimentSettings.info` is populated by a server-side (`utils/ssr/getExperimentInfo`). The call `getExperimentInfo` gets data from the `experiments` table. Right now, we don't have any record of `sampleIds` inside the `samples` table. Hence, the easiest way is to save `sampleIds` into `experiments` before running `GEM2S`.

This can be done, but to me it seems to be a "hacky" fix, which might cause problems down the line. Hence, I'd rather this problem be moved to another ticket and given more thought.

.github/workflows/ci.yaml
aerlaut: This was a change I made so that this branch compiled. It's already separated into PR #189

cosa65 (approved):

src/api/route-services/gem2s.js
```diff
@@ -76,7 +79,15 @@ class Gem2sService {
...
-    return taskParams;
+    const orderInvariantSampleIds = [...sampleIdsInOrder].sort();
+
+    const hashParams = {
+      sampleIds: orderInvariantSampleIds,
+      sampleNames: orderInvariantSampleIds.map((sampleId) => samples[sampleId].name),
+      metadata: taskParams.metadata,
+    };
+
+    return { taskParams, hashParams };
```
cosa65, line 92: I think we could add some unit testing to  this part at least to make sure that in the future we don't break this dependency on what causes it to run and what doesn't

cosa65 (changes requested):

cosa65 (approved):
Approved so you can merge tomorrow, just let me know if you feel like anotehr review would be necessary.

src/api/general-services/pipeline-manage/index.js
```diff
@@ -157,11 +155,7 @@ const createQCPipeline = async (experimentId, processingConfigUpdates) => {
   const accountId = await config.awsAccountIdPromise;
   const roleArn = `arn:aws:iam::${accountId}:role/state-machine-role-${config.clusterEnv}`;
   logger.log(`Fetching processing settings for ${experimentId}`);
-  const { processingConfig } = await experimentService.getProcessingConfig(experimentId);
-
-  const { samples } = await samplesService.getSamplesByExperimentId(experimentId);
-
-  const sampleIds = Object.keys(samples);
+  const { processingConfig, sampleIds } = await experimentService.getExperimentAttributes(experimentId, ['processingConfig', 'sampleIds']);
```
aerlaut: Create a wrapper function - for now - to make it uniform like the others.

src/api/route-services/experiment.js
```diff
@@ -70,7 +94,7 @@ class ExperimentService {
 
     const marshalledData = convertToDynamoDbRecord({
       ':experimentName': body.name,
-      ':createdAt': body.createdAt,
+      ':createdDate': body.createdDate,
```
aerlaut, line 77: We'll need to make a migration for this

aerlaut: An easier solution is to just copy `createdAt` to `createdDate` in the migration

tests/api/route-services/gem2s.test.js
```diff
@@ -183,4 +187,138 @@ describe('gem2sShouldRun', () => {
...
+        .mockImplementationOnce(() => Promise.resolve({ metadataKeys: [] })),
+    }));
+
+
+    const { taskParams: task2, hashParams: hash2 } = await Gem2sService.generateGem2sParams('experimentId');
+
+    expect(hash1).toEqual(hash2);
+  });
+
+  it('generateGem2sTaskParams returns different hash param if passed diferent information on vital fields', async () => {
```
aerlaut: Splint into 3 different tests

src/api/route-services/gem2s.js
aerlaut: Written

src/api/general-services/pipeline-manage/index.js
aerlaut: created

tests/api/route-services/gem2s.test.js
aerlaut: split

src/api/route-services/experiment.js
aerlaut: created, now testing

PR #199, opened 2021-08-09
Execution ARN does not exist in staging

kafkasl:
Returning NOT_CREATED status when the state machine execution ARN does not exist in staging

State machines in staging are removed after some time, in this situation we return NOT_CREATED status so that the pipeline can be run again

Staging link : https://ui-agi-api199.scp-staging.biomage.net/


aerlaut (approved):
Looks good. code tested and it works in this staging

PR #228, opened 2021-09-15
Fix project ids fetching

cosa65:
# Background
https://ui-martinfosco-api228.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [ ] Deployed to staging
- [ ] Passed integration tests 

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/api/route-services/projects.js
+++ b/src/api/route-services/projects.js
@@ -124,7 +124,7 @@ class ProjectsService {
   /**
    * Returns information about a group of projects.
    *
-   * @param {Array} projectIds An array of projectId values that are to be queried.
+   * @param {Array} projectIds A Array of projectId values that are to be queried.
    * @returns An object containing descriptions of projects.
    */
   async getProjectsFromIds(projectIds) {
@@ -136,7 +136,7 @@ class ProjectsService {
     const params = {
       RequestItems: {
         [this.tableName]: {
-          Keys: [...projectIds].map((projectUuid) => convertToDynamoDbRecord({ projectUuid })),
+          Keys: projectIds.map((projectUuid) => convertToDynamoDbRecord({ projectUuid })),
         },
       },
     };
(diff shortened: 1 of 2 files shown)


ivababukova (approved):

src/api/route-services/projects.js
```diff
@@ -118,16 +114,24 @@ class ProjectsService {
       projectIds = projectIds.concat(newProjectIds);
     }
 
-    return this.getProjectsFromIds(new Set(projectIds));
+    // Remove duplicates (when we support multi experiment projects
+    // we might have repeated projectId's)
+    projectIds = [...new Set(projectIds)];
```
ivababukova, line 119: why making it to list now instead of leaving it to be a set like before?

ivababukova: wouldn't making it a list afterwards add more complexity?

cosa65: because I don't see the reason why `getProjectsFromIds` should be receiving a set, it is working on it as if it were an array.

The only reason I can think of that we were making it into a set was to remove duplicates, so I think using this is more explicit about the fact that we want to remove duplicates and doesn't force other functions to adapt to a set when they don't need to.

cosa65: doing `[...new Set(projectIds)]` is a fairly common pattern to remove duplicates in javascript: https://stackoverflow.com/questions/9229645/remove-duplicate-values-from-js-array

ivababukova: Ok, I see. My original question was about why do we have to make them into a set originally, but now I understand (remove duplicate projectIds). I think that if `getProjectsFromIds` works with Set, I wouldn't be making it into an Array, I wasn't questioning that. Making it to Array adds one more operation, and I don't see the benefits of it. I would revert back to the previous code

cosa65: Resolved in a call

PR #231, opened 2021-09-17
[BIOMAGE-1433] Missing samples in projects

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1433

https://ui-martinfosco-ui515-api231.scp-staging.biomage.net/

https://github.com/biomage-ltd/ui/pull/515

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [ ] Deployed to staging
- [ ] Passed integration tests 

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/api/route-services/__mocks__/projects.js
+++ b/src/api/route-services/__mocks__/projects.js
@@ -1,6 +1,16 @@
 const { NotFoundError, OK } = require('../../../utils/responses');
 
-const mockCreateProject = jest.fn(() => new Promise((resolve) => {
+const mockCreateProject = jest.fn((projectUuid, body) => new Promise((resolve, reject) => {
+  if (!projectUuid
+    || !body
+    || !body.uuid
+  ) {
+    const err = new Error('Invalid body');
+    err.status = 400;
+
+    reject(err);
+  }
+
   resolve(OK());
 }));
 
--- a/src/api/route-services/__mocks__/projects.js
+++ b/src/api/route-services/__mocks__/projects.js
@@ -1,5 +1,9 @@
 const { NotFoundError, OK } = require('../../../utils/responses');
 
+const mockCreateProject = jest.fn(() => new Promise((resolve) => {
+  resolve(OK());
+}));
+
 const mockGetProject = jest.fn((projectUuid) => new Promise((resolve) => {
   if (projectUuid === 'unknown-project') {
     throw new NotFoundError('Project not found');
@@ -56,6 +60,7 @@ const mockDeleteProject = jest.fn((projectUuid) => new Promise((resolve) => {
 }));
 
 const mock = jest.fn().mockImplementation(() => ({
+  createProject: mockCreateProject,
   getExperiments: mockGetExperiments,
   getProject: mockGetProject,
   getProjects: mockGetProjects,
--- a/src/api/route-services/__mocks__/samples.js
+++ b/src/api/route-services/__mocks__/samples.js
@@ -70,11 +70,30 @@ const mockDeleteSamples = jest.fn((projectUuid, sampleUuid) => new Promise((reso
   resolve(OK());
 }));
 
+const mockAddSample = jest.fn(
+  (projectUuid, experimentId, body) => new Promise((resolve, reject) => {
+    if (!projectUuid
+      || !body
+      || !experimentId
+      || !body.projectUuid
+      || !body.uuid
+    ) {
+      const err = new Error('Invalid body');
+      err.status = 400;
+
+      reject(err);
+    }
+
+    resolve({ data: { message: 'sucess', code: 200 } });
+  }),
+);
+
 const mock = jest.fn().mockImplementation(() => ({
   getSamples: mockGetSamples,
   getSamplesByExperimentId: mockGetByExperimentId,
   updateSamples: mockUpdateSamples,
   deleteSamplesEntry: mockDeleteSamples,
+  addSample: mockAddSample,
 }));
 
 module.exports = mock;
--- a/src/api/route-services/projects.js
+++ b/src/api/route-services/projects.js
@@ -17,7 +17,7 @@ const experimentService = new ExperimentService();
 
 class ProjectsService {
   constructor() {
-    this.tableName = `projects-${config.clusterEnv}`;
+    this.projectsTableName = `projects-${config.clusterEnv}`;
     this.samplesTableName = `samples-${config.clusterEnv}`;
   }
 
@@ -28,7 +28,7 @@ class ProjectsService {
     });
 
     const params = {
-      TableName: this.tableName,
+      TableName: this.projectsTableName,
       Key: marshalledKey,
     };
 
@@ -67,7 +67,7 @@ class ProjectsService {
     });
 
     const params = {
-      TableName: this.tableName,
+      TableName: this.projectsTableName,
       Key: marshalledKey,
       UpdateExpression: 'SET projects = :project',
       ExpressionAttributeValues: marshalledData,
@@ -147,7 +147,7 @@ class ProjectsService {
     const dynamodb = createDynamoDbInstance();
     const params = {
       RequestItems: {
-        [this.tableName]: {
+        [this.projectsTableName]: {
           Keys: projectIds.map((projectUuid) => convertToDynamoDbRecord({ projectUuid })),
         },
       },
@@ -155,7 +155,7 @@ class ProjectsService {
 
     const data = await safeBatchGetItem(dynamodb, params);
 
-    const existingProjectIds = new Set(data.Responses[this.tableName].map((entry) => {
+    const existingProjectIds = new Set(data.Responses[this.projectsTableName].map((entry) => {
       const newData = convertToJsObject(entry);
       return newData.projects.uuid;
     }));
@@ -178,7 +178,7 @@ class ProjectsService {
         return newProject;
       });
 
-    data.Responses[this.tableName].forEach((entry) => {
+    data.Responses[this.projectsTableName].forEach((entry) => {
       const newData = convertToJsObject(entry);
       projects.push(newData.projects);
     });
@@ -192,7 +192,7 @@ class ProjectsService {
     const marshalledKey = convertToDynamoDbRecord({ projectUuid });
 
     const params = {
-      TableName: this.tableName,
+      TableName: this.projectsTableName,
       Key: marshalledKey,
     };
 
@@ -216,7 +216,7 @@ class ProjectsService {
     });
 
     const params = {
-      TableName: this.tableName,
+      TableName: this.projectsTableName,
       Key: marshalledKey,
     };
 
--- a/src/api/route-services/projects.js
+++ b/src/api/route-services/projects.js
@@ -71,11 +71,13 @@ class ProjectsService {
       Key: marshalledKey,
       UpdateExpression: 'SET projects = :project',
       ExpressionAttributeValues: marshalledData,
+      ReturnValues: 'UPDATED_NEW',
     };
 
     const dynamodb = createDynamoDbInstance();
 
-    await dynamodb.updateItem(params).send();
+    await dynamodb.updateItem(params).promise();
+
     return OK();
   }
 
@@ -233,7 +235,7 @@ class ProjectsService {
         await Promise.all(deletePromises);
       }
 
-      await dynamodb.deleteItem(params).send();
+      await dynamodb.deleteItem(params).promise();
 
       return OK();
     } catch (e) {
--- a/src/api/route-services/projects.js
+++ b/src/api/route-services/projects.js
@@ -62,6 +62,8 @@ class ProjectsService {
     const dynamodb = createDynamoDbInstance();
 
     try {
+      console.log('marshalledDataDebug');
+      console.log(JSON.stringify(marshalledData));
       await dynamodb.updateItem(params).send();
       return OK();
     } catch (e) {
--- a/src/api/route-services/projects.js
+++ b/src/api/route-services/projects.js
@@ -71,7 +71,6 @@ class ProjectsService {
       Key: marshalledKey,
       UpdateExpression: 'SET projects = :project',
       ExpressionAttributeValues: marshalledData,
-      ReturnValues: 'UPDATED_NEW',
     };
 
     const dynamodb = createDynamoDbInstance();
@@ -257,7 +256,6 @@ class ProjectsService {
       Key: convertToDynamoDbRecord({ experimentId }),
       UpdateExpression: 'SET samples = :emptySamples, projectUuid = :projectUuid',
       ExpressionAttributeValues: marshalledData,
-      ReturnValues: 'UPDATED_NEW',
     };
 
     await dynamodb.updateItem(emptyAttributeParams).promise();
--- a/src/api/route-services/projects.js
+++ b/src/api/route-services/projects.js
@@ -62,8 +62,6 @@ class ProjectsService {
     const dynamodb = createDynamoDbInstance();
 
     try {
-      console.log('marshalledDataDebug');
-      console.log(JSON.stringify(marshalledData));
       await dynamodb.updateItem(params).send();
       return OK();
     } catch (e) {
--- a/src/api/route-services/projects.js
+++ b/src/api/route-services/projects.js
@@ -18,6 +18,7 @@ const experimentService = new ExperimentService();
 class ProjectsService {
   constructor() {
     this.tableName = `projects-${config.clusterEnv}`;
+    this.samplesTableName = `samples-${config.clusterEnv}`;
   }
 
   async getProject(projectUuid) {
@@ -42,6 +43,19 @@ class ProjectsService {
     return response;
   }
 
+  async createProject(projectUuid, project) {
+    logger.log(`Creating project with id ${projectUuid}`);
+
+    const experimentId = project.experiments[0];
+
+    await Promise.all([
+      this.createEmptySamplesEntry(projectUuid, experimentId),
+      this.updateProject(projectUuid, project),
+    ]);
+
+    return OK();
+  }
+
   async updateProject(projectUuid, project) {
     logger.log(`Updating project with id ${projectUuid}`);
     const marshalledKey = convertToDynamoDbRecord({
@@ -61,13 +75,8 @@ class ProjectsService {
 
     const dynamodb = createDynamoDbInstance();
 
-    try {
-      await dynamodb.updateItem(params).send();
-      return OK();
-    } catch (e) {
-      if (e.statusCode === 400) throw new NotFoundError('Project not found');
-      throw e;
-    }
+    await dynamodb.updateItem(params).send();
+    return OK();
   }
 
   /**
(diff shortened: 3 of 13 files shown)


src/api/route-services/projects.js
```diff
@@ -61,13 +75,8 @@ class ProjectsService {
 
     const dynamodb = createDynamoDbInstance();
 
-    try {
-      await dynamodb.updateItem(params).send();
-      return OK();
-    } catch (e) {
-      if (e.statusCode === 400) throw new NotFoundError('Project not found');
```
cosa65, line 68: Removed the try catch because I think a 400 could mean many things, not necessarily that the project was not found, so this custom error might cause more confusion than be useful

src/api/routes/samples.js
```diff
@@ -2,31 +2,50 @@ const SamplesService = require('../route-services/samples');
 
 const samplesService = new SamplesService();
 
+const {
+  expressAuthorizationMiddleware,
```
cosa65, line 6: Added auth for the samples endpoints because it didn't have before (it seems that, before this, any user could call and run samples updates or get samples data of any project/experiment)

cosa65: There's also other endpoints without auth that we shoudl look into, let me know if you want me to work on that in this PR

cosa65:
Pending unit tests addition

src/api/routes/samples.js
```diff
@@ -2,31 +2,50 @@ const SamplesService = require('../route-services/samples');
...
+        .then((data) => res.json(data))
+        .catch(next);
+    }],
+  'samples#getSamplesByExperimentId': [
+    expressAuthorizationMiddleware, (req, res, next) => {
+      samplesService.getSamplesByExperimentId(req.params.experimentId)
+        .then((data) => res.json(data))
+        .catch(next);
+    }],
+  'samples#update': [
```
cosa65, line 23: Check if this endpint is stil lbeing used

cosa65: probably used for internal sampels proeprties updates

src/api/route-services/samples.js
```diff
@@ -144,6 +145,65 @@ class SamplesService {
...
+
+    const addSampleToSamples = async () => {
+      const dynamodb = createDynamoDbInstance();
+
+      const marshalledKey = convertToDynamoDbRecord({
+        experimentId,
+      });
+
+      const marshalledData = convertToDynamoDbRecord({
+        ':newSample': [newSample],
```
cosa65: Check that this should be an array

src/api/route-services/__mocks__/samples.js
```diff
@@ -70,11 +70,30 @@ const mockDeleteSamples = jest.fn((projectUuid, sampleUuid) => new Promise((reso
...
 
+const mockAddSample = jest.fn(
+  (projectUuid, experimentId, body) => new Promise((resolve, reject) => {
+    if (!projectUuid
+      || !body
+      || !experimentId
+      || !body.projectUuid
+      || !body.uuid
+    ) {
+      const err = new Error('Invalid body');
```
cosa65, line 81: This invalid is being received from inside the route-service (the mock), is this fine or should it be handled by the api validation?

src/api/routes/projects.js
```diff
@@ -7,7 +7,17 @@ const {
 
 const projectsService = new ProjectsService();
 
+console.log('projectsServiceDebug124e');
+console.log(projectsService);
```
aerlaut: Don't forget to remove this

src/api/routes/samples.js
aerlaut: Hmm.. I think it's better to create a new ticket. Don't want to drag this ticket too long.

src/api/routes/samples.js
aerlaut: Yea this is still used in the UI by `saveSamples`.

src/api/routes/samples.js
aerlaut: it does look weird though. Why don't we have the auth middleware for samples.

src/api/routes/samples.js
```diff
@@ -2,31 +2,50 @@ const SamplesService = require('../route-services/samples');
...
+    }],
+  'samples#add': [
+    expressAuthorizationMiddleware,
+    (req, res, next) => {
+      samplesService.addSample(req.params.projectUuid, req.params.experimentId, req.body)
+        .then((data) => res.json(data))
+        .catch(next);
+    },
+  ],
   'samples#uploadSampleFileUrl': (req, res, next) => {
```
aerlaut: not adding `authMiddleWare` here because? it's handled by S3?

src/api/route-services/projects.js
```diff
@@ -42,6 +43,19 @@ class ProjectsService {
     return response;
   }
 
+  async createProject(projectUuid, project) {
+    logger.log(`Creating project with id ${projectUuid}`);
+
+    const experimentId = project.experiments[0];
+
+    await Promise.all([
+      this.createEmptySamplesEntry(projectUuid, experimentId),
```
aerlaut, line 52: Why do we need to create an empty samples entry? If we upload files, can't the API just save it if there is no samples entry before?

src/api/route-services/projects.js
```diff
@@ -42,6 +43,19 @@ class ProjectsService {
     return response;
   }
 
+  async createProject(projectUuid, project) {
```
aerlaut, line 46: Is there any case where this function could fail? Should we handle it?

src/api/route-services/samples.js
```diff
@@ -144,6 +145,65 @@ class SamplesService {
     return OK();
   }
 
+  async addSample(projectUuid, experimentId, newSample) {
+    logger.log(`Adding samples ${JSON.stringify(newSample)} to project with id ${projectUuid}`);
```
aerlaut: Wouldn't doing `JSON.stringify` cause this log to be long? Is it not enough to just print the `sampleUuid`?

src/api/route-services/samples.js
```diff
@@ -144,6 +145,65 @@ class SamplesService {
...
+      const marshalledData = convertToDynamoDbRecord({
+        ':newSampleId': [newSample.uuid],
+      });
+
+      const params = {
+        TableName: this.projectsTableName,
+        Key: marshalledKey,
+        UpdateExpression: 'SET projects.samples = list_append(projects.samples, :newSampleId)',
+        ExpressionAttributeValues: marshalledData,
+        ReturnValues: 'UPDATED_NEW',
```
aerlaut: Do we need this if we're not using the return value?

src/api/route-services/samples.js
```diff
@@ -144,6 +145,65 @@ class SamplesService {
...
+
+      const params = {
+        TableName: this.samplesTableName,
+        Key: marshalledKey,
+        UpdateExpression: 'SET samples.#newSampleId = :newSample, projectUuid = :projectUuid',
+        ExpressionAttributeNames: {
+          '#newSampleId': newSample.uuid,
+        },
+        ExpressionAttributeValues: marshalledData,
+        ReturnValues: 'UPDATED_NEW',
```
aerlaut: Do we need this if we're not using the return value?

src/api/route-services/projects.js
```diff
@@ -224,14 +235,33 @@ class ProjectsService {
...
+      ':emptySamples': {},
+      ':projectUuid': projectUuid,
+    });
+
+    const emptyAttributeParams = {
+      TableName: this.samplesTableName,
+      Key: convertToDynamoDbRecord({ experimentId }),
+      UpdateExpression: 'SET samples = :emptySamples, projectUuid = :projectUuid',
+      ExpressionAttributeValues: marshalledData,
+      ReturnValues: 'UPDATED_NEW',
```
aerlaut: do we need this if we're not using the return value?

src/api/route-services/projects.js
```diff
@@ -57,17 +71,14 @@ class ProjectsService {
       Key: marshalledKey,
       UpdateExpression: 'SET projects = :project',
       ExpressionAttributeValues: marshalledData,
+      ReturnValues: 'UPDATED_NEW',
```
aerlaut: do we need this if we're not using the return value?

src/api/route-services/projects.js
```diff
@@ -18,6 +18,7 @@ const experimentService = new ExperimentService();
 class ProjectsService {
   constructor() {
     this.tableName = `projects-${config.clusterEnv}`;
```
aerlaut: to make it more obvious, would it be better to rename this to `projectsTableName`?

src/api/route-services/__mocks__/samples.js
aerlaut: I think it would be better if it can be handled by the API validation. It can be detected earlier and so we don't have to code a handler inside the API to take care of it. I added one yesterday for `POST /runGEm2s`. It requires `paramsHash` to be available in the body.

aerlaut: Moving it there would also allow us to have a better mock for `mockAddSample`, I think.

src/specs/api.yaml
```diff
@@ -756,14 +756,65 @@ paths:
...
+    
     put:
       summary: Update project
       operationId: updateProject
       x-eov-operation-id: 'projects#update'
       x-eov-operation-handler: routes/projects
       responses:
         '200':
-          description: Update sample object
+          description: Update project object
```
aerlaut, line 830: Nice catch 👍

tests/api/route-services/projects.test.js
```diff
@@ -404,4 +405,33 @@ describe('tests for the projects service', () => {
...
+
+  test('createProject creates entries in projects and samples correctly', (done) => {
+    const updateItemSpy = mockDynamoUpdateItem({});
+
+    const expectedSamplesUpdate = {
+      TableName: 'samples-test',
+      Key: AWS.DynamoDB.Converter.marshall({ experimentId: 'experiment-1' }),
+      UpdateExpression: 'SET samples = :emptySamples, projectUuid = :projectUuid',
+      ExpressionAttributeValues: AWS.DynamoDB.Converter.marshall({ ':emptySamples': {}, ':projectUuid': 'project-1' }),
+      ReturnValues: 'UPDATED_NEW',
```
aerlaut: Don't forget to remove these `UPDATED_NEW` if they are not used anymore,

tests/api/route-services/projects.test.js
```diff
@@ -404,4 +405,33 @@ describe('tests for the projects service', () => {
...
+    };
+
+    (new ProjectsService()).createProject('project-1', mockProject)
+      .then((res) => {
+        expect(res).toEqual(OK());
+
+        expect(updateItemSpy.mock.calls).toContainEqual([expectedSamplesUpdate]);
+        expect(updateItemSpy.mock.calls).toContainEqual([expectedProjectsUpdate]);
+      })
+      .then(() => done());
```
aerlaut: Rather than doing this, can you make this test to be `async`, and then break this piece of code into `awaits`? I think it' easier to read,

tests/api/route-services/samples.test.js
```diff
@@ -149,4 +148,58 @@ describe('tests for the samples service', () => {
...
+
+    (new SamplesService()).addSample('projectUuid', 'experimentId', mockSampleToAdd)
+      .then((res) => {
+        expect(res).toEqual(OK());
+
+        expect(updateItemSpy.mock.calls).toContainEqual([expectedSamplesUpdate]);
+        expect(updateItemSpy.mock.calls).toContainEqual([expectedProjectsUpdate]);
+      })
+      .then(() => done());
+  });
```
aerlaut: Likewise as fo the projects, if possible change this test function body into `async` and break this code open.

tests/api/route-services/samples.test.js
```diff
@@ -149,4 +148,58 @@ describe('tests for the samples service', () => {
       })
       .then(() => done());
   });
+
+  it('addSample works', (done) => {
```
aerlaut: Is there any case where `addSample` doesn't work? Are we handling it? Do we need to test for it?

src/api/route-services/projects.js
cosa65: If we are doing an update where we just overwite the existing then it is fine.

In our case we are updating an existing element, and creating it empty if it doesn't exist can't be done in the same call that does an update to a nested property (it is actually a problem I already ran into with the processing config back when we didn't create it originally)

tests/api/route-services/projects.test.js
cosa65: Ok, should I do this with all the other tests? Since they all have this pattern

src/api/route-services/samples.js
```diff
@@ -144,6 +145,65 @@ class SamplesService {
...
+
+      const marshalledData = convertToDynamoDbRecord({
+        ':newSample': newSample,
+        ':projectUuid': projectUuid,
+      });
+
+      const params = {
+        TableName: this.samplesTableName,
+        Key: marshalledKey,
+        UpdateExpression: 'SET samples.#newSampleId = :newSample, projectUuid = :projectUuid',
```
cosa65: check if this is necessary: `projectUuid`

src/api/routes/samples.js
cosa65: Ok, then we need to add a ticket for this

src/api/route-services/projects.js
cosa65: Dynamodb has[ a lot of error codes](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/CommonErrors.html) por possible failures, do you mean we should begin handling dynamo errors?

If you think we should go that way I guess we could put all dynamo requests in a try catch, in the catch log the error and then rethrow (or throw hiding the original error).

However, I feel like that is something we should probably add in a different ticket since it's something we are not really doing anywhere (handling dynamo errors in a comprehensive way).

tests/api/route-services/samples.test.js
cosa65: See https://github.com/biomage-ltd/api/pull/231#discussion_r722684737

src/api/route-services/projects.js
ivababukova: Did you consider leaving the try/catch, but removing the if statement, so that it catches, logs and returns any dynamoDB error that can occur?

src/api/route-services/projects.js
```diff
@@ -224,14 +234,32 @@ class ProjectsService {
...
+
+    const marshalledData = convertToDynamoDbRecord({
+      ':emptySamples': {},
+      ':projectUuid': projectUuid,
+    });
+
+    const emptyAttributeParams = {
+      TableName: this.samplesTableName,
+      Key: convertToDynamoDbRecord({ experimentId }),
+      UpdateExpression: 'SET samples = :emptySamples, projectUuid = :projectUuid',
```
ivababukova, line 257: did you check if we still need to have `projectUuid = :projectUuid`?

tests/api/routes/projects.test.js
```diff
@@ -54,6 +54,53 @@ describe('tests for projects route', () => {
...
+      lastModified: '',
+      uuid: 'project-1',
+      experiments: [],
+      lastAnalyzed: null,
+    };
+
+    request(app)
+      .post('/v1/projects/someId')
+      .send(payload)
+      .expect(200)
```
ivababukova: why expect 200 here? Aren't you expecting 400?

tests/api/routes/projects.test.js
```diff
@@ -54,6 +54,53 @@ describe('tests for projects route', () => {
...
+
+  test('Creating project with invalid shape project returns 400', async (done) => {
+    const payload = {
+      name: 'Test project',
+      description: '',
+      createdDate: '',
+      lastModified: '',
+      uuid: 'project-1',
+      experiments: [],
+      lastAnalyzed: null,
```
ivababukova: This looks quite valid. Maybe you forgot to make it invalid and modify the 200 into 400?

src/api/route-services/projects.js
cosa65: so catching, logging and throwing again? ok, sounds good

src/api/route-services/projects.js
cosa65: yes, we didn't and I removed it, but it isn't here that I did that, this particular update we do need it because we are creating the entry from scratch

cosa65: it's `async addSample(projectUuid, experimentId, newSample) {` in `samples.js`

tests/api/routes/projects.test.js
cosa65: that's what happened, I was looking into the mocks yet, they are returning mocked invalids which don't really help to test the api validation in any way (it should be caught before it reaches the mock by the schema validator but the vlaidation isn't really restrictive)

cosa65: I think I will fix these validation bodies for samples and projects endpoints along with these tests but in a different branch since it is marginally related to the original purpose of this PR and I don't want to keep making it larger. 

The new PR will be to fix the tests in projects and samples and also the api validation

src/api/route-services/projects.js
cosa65: I'll create a ticket for this, otherwise it will bloat this PR

src/api/route-services/__mocks__/samples.js
cosa65: Yes, I'll add a new ticket for doing this for all the mocks (all of them have this pattern) and also for fixing the api validation

tests/api/routes/projects.test.js
cosa65: see https://github.com/biomage-ltd/api/pull/231#discussion_r723576946

src/api/route-services/__mocks__/projects.js
```diff
@@ -1,5 +1,19 @@
...
+const mockCreateProject = jest.fn((projectUuid, body) => new Promise((resolve, reject) => {
+  if (!projectUuid
+    || !body
+    || !body.uuid
+  ) {
+    const err = new Error('Invalid body');
+    err.status = 400;
+
+    reject(err);
+  }
```
aerlaut, line 12: So have you created a new ticket for this? Can you paste the link here?

tests/api/route-services/projects.test.js
```diff
@@ -129,16 +128,14 @@ describe('tests for the projects service', () => {
...
-        expect(projectService.getProjectsFromIds).toHaveBeenCalledWith(expectedResult);
-      })
-      .then(() => done());
+    expect(res).toEqual(expectedResult);
+
+    expect(projectService.getProjectsFromIds).toHaveBeenCalledWith(expectedResult);
   });
 
-  it('GetProjects gets all the projects across many pages of scan results when the first page is empty', async (done) => {
+  test('GetProjects gets all the projects across many pages of scan results when the first page is empty', async () => {
```
aerlaut, line 138: why do you change to `test`? some cases are sill using `it`

src/api/route-services/__mocks__/projects.js
cosa65: https://biomage.atlassian.net/browse/BIOMAGE-1496

tests/api/route-services/projects.test.js
cosa65: `test` is an alias of `it`, they both do the same, I use them because it fits syntactically with the name of the test

cosa65: "it gets all the projects"
"test getProjects gets all the projects"

aerlaut (approved):

PR #236, opened 2021-09-28
Fix lack of selfLink on modern versions of Kubernetes

marcellp:
This is a hot-fix that fixes the bug reported by Oliver [here](https://this-is-biomage.slack.com/archives/C0250TDU1PF/p1632776099003200).

`selfLink` as a metadata value is deprecated in k8s 1.20 and removed in 1.21. See https://github.com/kubernetes/enhancements/issues/1164.


ivababukova (approved):
if it works in staging, let's merge it.

PR #244, opened 2021-10-14
[BIOMGE-1057] returning mocked success in production when executionARN does not exist

kafkasl:
Ticket: https://biomage.atlassian.net/browse/BIOMAGE-1517
Staging: https://ui-flaky-silver-gibbon.scp-staging.biomage.net/data-management

Main reason

aws deletes state machines after 90 days so backend-status calls fail. This PR is a temporary fix to avoid the "It's not you it's us" error. As discussed with Vicky, we will assume the user already run their experiment to avoid forcing them to rerun and lose their settings. The next fix will check whether those files exist in S3 to avoid redirecting users who actually haven't run the pipeline.

Changes
* Moved hookRunner to /hooks folder
* Added many tests to pipeline-status
* Refactored pipeline-status response handling to avoid forgetting to add params in some code paths

### DIFF
--- a/src/api/general-services/pipeline-status.js
+++ b/src/api/general-services/pipeline-status.js
@@ -22,8 +22,8 @@ const notCreatedStatus = {
   completedSteps: [],
 };
 
-const date = (new Date()).toISOString();
-const mockedCompletedStatus = {
+// const date = (new Date()).toISOString();
+const buildCompletedStatus = (date) => ({
   qc: {
     startDate: date,
     stopDate: date,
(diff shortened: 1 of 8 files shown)


src/api/general-services/pipeline-status.js
```diff
@@ -180,74 +241,54 @@ const getPipelineStatus = async (experimentId, processName) => {
...
+  //     status: execution.status,
+  //     error,
+  //     completedSteps,
+  //   },
+  // };
+
+  // if (processName === pipelineConstants.GEM2S_PROCESS_NAME) {
+  //   const { paramsHash } = pipelinesHandles[pipelineConstants.GEM2S_PROCESS_NAME];
+  //   response[pipelineConstants.GEM2S_PROCESS_NAME].paramsHash = paramsHash;
+  // }
```
aerlaut: Dont' forget to remove this

aerlaut (approved):
LGTM 👍

PR #294, opened 2022-01-28
Sharing exps

kafkasl:
# Description
Adds the new permissions mechanism but still relies on rbac_can_write to allow / deny permissions.

# Details

#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1647

https://ui-pol-api294.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/iac/pull/248

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api/route-services/access.js
```diff
@@ -0,0 +1,104 @@
...
+} = require('../../utils/dynamoDb');
+const { isRoleAuthorized } = require('./roles');
+
+
+const logger = getLogger('[AccessService] - ');
+
+class AccessService {
+  constructor() {
+    this.userAccessTableName = `user-access-${config.clusterEnv}`;
+    this.inviteAccessTableName = `invite-access-${config.clusterEnv}`;
```
ivababukova: better to put the particular database names in default-config and import the names from there

src/api/route-services/access.js
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    const params = {
+      TableName: this.userAccessTableName,
+      Key: key,
+    };
+
+    console.log(params);
+
+    const dynamodb = createDynamoDbInstance();
```
ivababukova: we don't need to create a new instance of dynamodb for every particular class function -- we can make `dynamodb` a class variable

kafkasl: I agree and that is what I tried to do. However, the way we mock AWS then does not work (because the instance is created when we import it). I considered whether refactoring this but I thought it should not be part of this PR because all of the route-services do this (creating an instance every time).

src/api/route-services/access.js
kafkasl: I'll create a ticket to move the names into a default-config

README.md
```diff
@@ -3,7 +3,7 @@ API
 ======
 
 
-The API of Cellscope (the Biomage single cell analysis platform).
+The API of Cellenics (the Biomage single cell analysis platform).
```
ivababukova: When you merge in the master branch, can you accept the changes to the readme in master? I fixed all these references to Cellscope and removed the "biomage single cell analysis platform" bits

src/utils/authMiddlewares.js
```diff
@@ -230,6 +232,40 @@ const authorize = async (authResource, claim, authByExperiment = true) => {
...
+
+/**
+ * General authorization middleware. Resolves with nothing on
+ * successful authorization, or an exception on unauthorized access.
+ *
+ * @param {*} userId The ID of the user to authorize.
+ * @param {*} resource The resource the user is requesting (either the URL or 'sockets').
+ * @param {*} method The HTTP method of the request (or null in the case of sockets)
+ * @param {*} authResource Either the experimentId or the projectUuid
+ * @param {*} authByExperiment if true => authResource is an experimentId, false => projectUuid
```
ivababukova, line 244: do we ever use the `authExperiment = false`? In what cases would we use it?

kafkasl: yes, there are some cases where we send the `projectUuid` instead of the experiment ID. I'd need to check which of them are specifically though

README.md
kafkasl: I merged it & re-pushed

src/api/route-services/access/index.js
```diff
@@ -0,0 +1,100 @@
...
+        userId,
+        experimentId,
+        projectId,
+        role,
+      },
+    };
+
+    const docClient = new AWS.DynamoDB.DocumentClient({
+      region: config.awsRegion,
+    });
```
ivababukova, line 40: this `DocumentClient` looks much more convenient than having to marshall and unmarshall data

ivababukova: can we use it for other methods too?

src/api/route-services/access/index.js
```diff
@@ -0,0 +1,100 @@
...
+
+    await docClient.put(params).promise();
+  }
+
+  async canAccessExperiment(userId, experimentId, url, method) {
+    logger.log(`GET experiment permissions ${experimentId} data`);
+    const key = convertToDynamoDbRecord({
+      experimentId,
```
ivababukova, line 48: what will happen in the `canAccessExperiment` method if we pass in nonexistent `experimentId` or nonexistent `userId`?

src/api/route-services/access/index.js
kafkasl: yes and no, I read that it has issues with paginations in some cases. considering we are going to remove this code (when we move to SQL) I did not spend more time making it consistent

src/api/route-services/access/index.js
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    const params = {
+      TableName: this.userAccessTableName,
+      Key: key,
+    };
+
+    const db = createDynamoDbInstance();
+    const response = await db.getItem(params).promise();
+    const { role } = convertToJsObject(response.Item);
```
ivababukova: if `response` is none, this line will error out. Can we add some error handling here and make `canAccessExperiment` always return false, unless the user and the experiment exist and the user has access to it?

src/api/route-services/access/index.js
kafkasl: then `const { role } = convertToJsObject(response.Item);` will return `undefined` and `isRoleAuthorized(role, url, method)` will correctly reject the authorization

src/api/route-services/access/index.js
kafkasl: no it will not, it will be undefined and be handled correctly by `isRoleAuthorized(...)`

src/api/route-services/access/roles.js
```diff
@@ -0,0 +1,46 @@
...
+};
+const isRoleAuthorized = (role, resource, method) => {
+  // if no valid role is provided, return not authorized
+  if (!ROLE_NAMES.includes(role)) {
+    return false;
+  }
+
+  // all roles have view permissions by default
+  if (method === 'GET') {
+    return true;
```
ivababukova, line 32: what would this return if a user tries to view an experiment that's not theirs?

ivababukova: Or would this case be ruled out by the `canAccessExperiment` function?

kafkasl: that's not possible because the `role` is obtained using the `experimentId` and `userId`

src/api/route-services/access/roles.js
```diff
@@ -0,0 +1,46 @@
...
+  // the roles' allowed resources match the requested resource
+  if (allowedResources[role].some((regex) => resource.match(regex))) {
+    return true;
+  }
+
+  return false;
+};
+
+module.exports = {
+  ROLE_NAMES, OWNER, ADMIN, EXPLORER, VIEWER, isRoleAuthorized,
```
ivababukova, line 45: If you export `ROLE_NAMES` and `OWNER, ADMIN, EXPLORER, VIEWER` are inside `ROLE_NAMES`, then why do you need to also export `OWNER, ADMIN, EXPLORER, VIEWER`?

src/api/route-services/access/index.js
ivababukova: so `response` will always have `Item` in it, even if it is empty?

src/api/route-services/access/index.js
ivababukova: wouldn't it be better if we make it so that if `role` is undefined, we reject the authorization directly in `canAccessExperiment` before we've called `isRoleAuthorized`? We know that having `undefined` for role will never be valid, why continuing further?

src/api/route-services/access/roles.js
kafkasl: `ROLE_NAMES` is just exported for the tests I can remove it. My issue is that if I put them in an object I can't access them by name (well I can, but the syntax / linter won't pick up reference errors)

src/api/route-services/projects.js
```diff
@@ -203,13 +215,6 @@ class ProjectsService {
 
       const experiments = await experimentService.getListOfExperiments(result.projects.experiments);
 
-      if (!withWritePermissions) {
-        experiments.forEach((experiment) => {
-          // eslint-disable-next-line no-param-reassign
-          delete experiment.rbac_can_write;
-        });
-      }
```
ivababukova, line 211: this is so weird, so we were deleting experiments while getting experiments?

src/utils/authMiddlewares.js
ivababukova: and I imagine that this will disappear once we finish with the Data Model?

src/api/route-services/access/index.js
ivababukova: makes sense!

src/api/route-services/access/roles.js
ivababukova: I see. Not a problem exporting them then

src/api/route-services/projects.js
kafkasl: we were deleting the field `rbac_can_write` in order to make the experiment "read-only" but this is not used at all and I am not sure it ever has.

src/api/route-services/access/index.js
kafkasl: no, when there is no Item the call `convertToJsObject(undefined)` will result in `role = undefined` which will be correctly handled by `isRoleAuthorized`. However, I've refactored to make explicit the handling of the `undefined` and avoid confusion in the future.

src/api/route-services/access/index.js
kafkasl: yes, I did something like that

src/utils/authMiddlewares.js
kafkasl: you imagine correctly!

ivababukova (approved):

PR #300, opened 2022-02-15
added delete permissions on delete experiment

kafkasl:
# Description
Added
* Deleting permissions table entries when deleting an experiment
* Changed wrong `adminArn` for staging in config file.
* `createdDate` field for permissions table

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1647

https://ui-stealthy-tomato-mongoose.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates
- [x] Relevant Github READMEs updated **or** no GitHub README updates required.
- [x] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional

### DIFF
--- a/src/api/route-services/access/index.js
+++ b/src/api/route-services/access/index.js
@@ -24,7 +24,6 @@ class AccessService {
   // grantRole adds or replaces role for userId and experimentId
   async grantRole(userId, experimentId, projectId, role) {
     logger.log(`PUT experiment role ${experimentId} data`);
-
     const params = {
       TableName: this.userAccessTableName,
       Item: {
@@ -32,6 +31,7 @@ class AccessService {
         experimentId,
         projectId,
         role,
+        createdDate: new Date().toISOString(),
       },
     };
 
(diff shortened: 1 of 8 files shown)


src/api/route-services/access/index.js
```diff
@@ -99,6 +99,74 @@ class AccessService {
...
+      },
+    };
+
+    await docClient.batchWrite(deleteParams).promise();
+  }
+
+  async deleteByExperimentId(table, experimentId) {
+    const items = await this.getExperimentEntries(table, experimentId);
+    if (items.length === 0) {
+      console.log(`no permissions rows found in ${table} for ${experimentId}`);
```
ivababukova: why don't you use the logger for this message?

src/api/route-services/access/index.js
```diff
@@ -99,6 +99,74 @@ class AccessService {
...
+    });
+    const getParams = {
+      TableName: table,
+      KeyConditionExpression: 'experimentId = :hkey',
+      ExpressionAttributeValues: {
+        ':hkey': experimentId,
+      },
+    };
+
+    const response = await docClient.query(getParams).promise();
```
ivababukova, line 116: do we need to worry about pagination here?

src/api/route-services/projects.js
```diff
@@ -243,6 +243,7 @@ class ProjectsService {
         const deletePromises = experiments.reduce((acc, experimentId) => {
           acc.push(experimentService.deleteExperiment(experimentId));
           acc.push(samplesService.deleteSamplesEntry(projectUuid, experimentId, sampleUuids));
+          acc.push(accessService.deleteExperiment(experimentId));
```
ivababukova: isn't it better if `accessService.deleteExperiment(experimentId)` is getting called as part of `experimentService.deleteExperiment(experimentId)` inside the `deleteExperiment` function of the `ExperimentService` class in `api/route-services/experiment`?

My reasoning is that it is part of deleting an experiment and if we call `accessService.deleteExperiment` from within `experimentService.deleteExperiment`, there are less moving parts involved. In particular, there is lower chance of us just calling `experimentService.deleteExperiment` from somewhere and forgetting to also call `accessService.deleteExperiment`

src/api/route-services/projects.js
```diff
@@ -243,6 +243,7 @@ class ProjectsService {
         const deletePromises = experiments.reduce((acc, experimentId) => {
           acc.push(experimentService.deleteExperiment(experimentId));
           acc.push(samplesService.deleteSamplesEntry(projectUuid, experimentId, sampleUuids));
```
ivababukova: I can't see a unit test to check that when we call `deleteProject` in `src/api/route-services/projects.js`, then the access gets deleted too. Could you modify the deleteProject test that we have right now to check for that?

src/api/route-services/projects.js
kafkasl: mmm I kind of agree, but then we should also move the `samplesService` delete inside the other one. This three calls will be removed and become a single one when we migrate into SQL anyway.

src/api/route-services/access/index.js
kafkasl: no, query returns up to 1Mb of data per partition key. We won't have more than a few rows per experimentID and anyway this will go away with SQL

src/api/route-services/access/index.js
kafkasl: I forgot to delete this

src/api/route-services/projects.js
ivababukova: ok, so if they will be removed as part of the data model migration, I agree it is not worth refactoring this

src/api/route-services/projects.js
kafkasl: I added checks for the 3 things as they were actually not checked.

ivababukova (approved):

PR #303, opened 2022-02-21
Set up client to connect to postgres

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1705

https://ui-martinfosco-api303.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/serviceConnections/createSQLClient.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const token = await signer.getAuthToken();
+
+  return {
+    host: endpoint,
+    port: 5432,
+    user: username,
+    password: token,
+    database: 'aurora_db',
+    ssl: { rejectUnauthorized: false },
```
kafkasl: I would add the DB configuration files in the `default-config.js` in an object called `config` so you can just
do `user: config.db.username` for example and you don't need to have a different function for `getLocalhostParams` (except maybe the endpoint)

I have also seen people using a knex config file, which can hold the values for all the different envs: https://gist.github.com/kkemple/9b810e98dd9caa669580 and https://blog.shahednasser.com/knex-js-tutorial-for-beginners/.

src/serviceConnections/createSQLClient.js
```diff
@@ -0,0 +1,75 @@
...
+
+const getLocalhostParams = () => ({
+  host: '127.0.0.1',
+  port: 5432,
+  user: 'api_role',
+  password: 'postgres', // pragma: allowlist secret
+  database: 'aurora_db',
+});
+
+const createSQLClient = async () => {
```
kafkasl: I am not sure if this is the right way of initializing knext. The docs state that:

> Initializing the library should normally only ever happen once in your application, as it creates a connection pool for the current database, you should use the instance returned from the initialize call throughout your library.

Doing this in a function will create a new pool connection every time if I am not wrong.

I have found a good sample project on how to structure knex and SQL stuff, I suggest we follow it for our implementation:
 https://github.dev/robmclarty/knex-express-project-sample/tree/main/server

src/serviceConnections/createSQLClient.js
cosa65: I think this fits better in its own config file rather than in default-config.js. For example, redis has its own folder with its own config.

I created the knexfile, when you can tell me your opinion on whether this would work.

src/SQLClient/index.js
```diff
@@ -0,0 +1,17 @@
+const knex = require('knex');
```
kafkasl: I was reading that calling fiiles `index.js` is not a good idea because it's too generic and makes it harder to search files by name and that they should match the exports so I think this should be called `SQLClient.js` and the folder `db`

src/SQLClient/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+const SQLClient = (() => {
+  let instance;
```
kafkasl, line 4: sholdn't  this `instance` be out of the `SQLClient` ? not sure how this scoping would work

kafkasl (approved):

PR #305, opened 2022-02-22
[BIOMAGE-1745] - Fix file upload error

aerlaut:
# Description
This PR and the associated UI PR introduces some improvements to make creation of sample records much more robust. Specifically this PR : 
- Saves the new sample id into the `experiments` table together with the insertion of the new sample id to the `samples` and `projects` table.
- Patches `samples` in the `samples` table instead of replacing the data, reducing the chance of error.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1745

https://ui-agi-ui638-api305.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/638

#### Integration test branch
master

# Merge checklist

### Code updates
- [X] Migrated any selector / reducer used to the new format.

### Manual/unit testing
- [X] Tested changes using InfraMock locally **or** no tests required for change, e.g. Kubernetes chart updates.
- [X] Validated that current unit tests for code work as expected and are sufficient for code coverage **or** no unit tests required for change, e.g. documentation update.
- [X] Unit tests written **or** no unit tests required for change, e.g. documentation update.
- 
### Integration testing

### Documentation updates
- [X] Relevant Github READMEs updated **or** no GitHub README updates required.
- [X] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional

### DIFF
--- a/src/api/route-services/samples.js
+++ b/src/api/route-services/samples.js
@@ -17,6 +17,7 @@ const logger = getLogger();
 
 class SamplesService {
   constructor() {
+    this.experimentsTableName = `experiments-${config.clusterEnv}`;
     this.samplesTableName = `samples-${config.clusterEnv}`;
     this.sampleFilesBucketName = `biomage-originals-${config.clusterEnv}`;
     this.projectsTableName = `projects-${config.clusterEnv}`;
@@ -204,9 +205,31 @@ class SamplesService {
       await dynamodb.updateItem(params).promise();
     };
 
+    const addSampleIdToExperiments = async () => {
+      const dynamodb = createDynamoDbInstance();
+
+      const marshalledKey = convertToDynamoDbRecord({
+        experimentId,
+      });
+
+      const marshalledData = convertToDynamoDbRecord({
+        ':newSampleId': [newSample.uuid],
+      });
+
+      const params = {
+        TableName: this.experimentsTableName,
+        Key: marshalledKey,
+        UpdateExpression: 'SET sampleIds = list_append(sampleIds, :newSampleId)',
+        ExpressionAttributeValues: marshalledData,
+      };
+
+      await dynamodb.updateItem(params).promise();
+    };
+
     await Promise.all([
       addSampleIdToProjects(),
       addSampleToSamples(),
+      addSampleIdToExperiments(),
     ]);
 
     return OK();
--- a/src/api/route-services/samples.js
+++ b/src/api/route-services/samples.js
@@ -81,27 +81,39 @@ class SamplesService {
   async updateSamples(projectUuid, experimentId, body) {
     logger.log(`Updating samples for project ${projectUuid} and expId ${experimentId}`);
 
+    const dynamodb = createDynamoDbInstance();
+
     const marshalledKey = convertToDynamoDbRecord({
       experimentId,
     });
 
+    const getSamplesParams = {
+      TableName: this.samplesTableName,
+      Key: marshalledKey,
+      ProjectionExpression: 'samples',
+    };
+
+    // Merge samples record with the body
+    const response = await dynamodb.getItem(getSamplesParams).promise();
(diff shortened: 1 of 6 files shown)


src/api/route-services/samples.js
```diff
@@ -81,27 +82,39 @@ class SamplesService {
...
 
+    const getSamplesParams = {
+      TableName: this.samplesTableName,
+      Key: marshalledKey,
+      ProjectionExpression: 'samples',
+    };
+
+    // Merge samples record with the body
+    const response = await dynamodb.getItem(getSamplesParams).promise();
+    const { samples } = convertToJsObject(response.Item);
```
cosa65, line 99: Do you think it would be a lot of work to, instead of getting, merging and setting it back in dynamo, send one request that replaces the samples that were updated? 

I think that would require some changes to the endpoints too though, so perhaps the motivation for doing it this way was to just avoid the racing conditions but not perform a full fix so that we don't spend too much time on the old endpoints we are going to stop using soon? Or was there another reason?

aerlaut: Unless there is an easy way to merge and update the sample cells with a DynamoDB UpdateExpression, I still think we'd need to fetch the samples, merge and insert the value back into the table.

ivababukova: I would be wary of changing this too much if this is some dynamodb specific changes, because we will soon be out of dynamo

src/api/route-services/samples.js
```diff
@@ -192,9 +205,31 @@ class SamplesService {
...
+      const params = {
+        TableName: this.experimentsTableName,
+        Key: marshalledKey,
+        UpdateExpression: 'SET sampleIds = list_append(sampleIds, :newSampleId)',
+        ExpressionAttributeValues: marshalledData,
+      };
+
+      await dynamodb.updateItem(params).promise();
+    };
+
```
ivababukova, line 228: I am surprised that we have to write this piece of code from scratch -- where in the api do we handle adding sample ids in the experiments table?

aerlaut: We don't. What we do to insert new sample ids to experiment is insert it into `experiments.sampleIds` redux and then save the whole updated `experiments` object.

src/api/route-services/samples.js
```diff
@@ -192,9 +205,31 @@ class SamplesService {
...
+      });
+
+      const params = {
+        TableName: this.experimentsTableName,
+        Key: marshalledKey,
+        UpdateExpression: 'SET sampleIds = list_append(sampleIds, :newSampleId)',
+        ExpressionAttributeValues: marshalledData,
+      };
+
+      await dynamodb.updateItem(params).promise();
```
ivababukova, line 226: some errors can happen here, better to surround it into try-catch block

src/api/route-services/samples.js
aerlaut: I agree. I'd say it's an acceptable solution for now.

src/api/route-services/samples.js
aerlaut: I think it's fine to do this. If there is an error, it'll cause the `Promise.all` block on [line 229](https://github.com/hms-dbmi-cellenics/api/blob/fda9a046e70e1e75f39f1c548065ba76ce805d62/src/api/route-services/samples.js#L229)   to throw an error, which will be received by Express router and returned as an error to the UI. If we want to catch the error and do some clean up (e.g. rollback by deleting the newly created sample id from `projects`, `experiments`, and `samples` table), I think it's best to have a try-catch on the `Promise.all` block. What do you think?

ivababukova: Looking at the UI PR, the UI counts on having this data in dynamo (the sample ids) even if the upload didn't fully work out. In this case, I wouldn't be doing a cleanup

ivababukova (approved):

PR #317, opened 2022-03-08
Add api v2 basic structure

cosa65:
# Description
This implementation is based on this [example](https://github.com/robmclarty/knex-express-project-sample/tree/main/server), with some small differences.

There are 3 main directories inside api.v2:
- **routes**: Similar to our previous implementation, just calls the controller corresponding for that route and does next() on a caught error so we get the error handler to log the errors.
- **controllers**: 1-to-1 relation with the routes, each route has a controller, it interacts with the model to perform updates.
- **model**: in its most simple interpretation: 1 for each table, but it might involve interaction with more than one table (for example when a join is required to get the sample ids that belong to an experiment), it adds a layer of abstraction from the interaction with sql. 
    The basic sql operations for experiment/user_access insertion are provided by `modelGuts` so we didn't need any custom model functions, but any table-specific queries should be implemented here. [This would be an example of a model file with some custom functions](https://github.com/robmclarty/knex-express-project-sample/blob/main/server/models/user.js). In my opinion for S3 related things `model` could be reused as well (it's may not necessarily be the best option so we can discuss it or see what works better when we are trying to implement an endpoint that affects s3 directly).

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1783

https://ui-martinfosco-ui652-api317.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/652

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api.v2/routes/health.js
```diff
@@ -0,0 +1,14 @@
...
+
+module.exports = {
+  'health#check': (req, res) => {
+    res.json({
+      status: 'up',
+      env: process.env.NODE_ENV,
+      clusterEnv: config.clusterEnv,
+      caching: CacheSingleton.get().isReady(),
+    });
+  },
```
cosa65, line 13: This doesn't follow the pattern of `route/` `controller/` `model/` but at the same time this handler in particular  is pretty simple, should I make it follow the pattern anyways or is it fine for very simple endpoints to do this? (or perhaps to only have the controller part).

src/specs/api.v2.yaml
```diff
@@ -0,0 +1,96 @@
+openapi: 3.0.0
+info:
+  title: API
+  version: 1.0.0
```
aerlaut: Update the version to 2.0.0?

src/specs/models/experiment-bodies/CreateExperiment.v2.yaml
```diff
@@ -0,0 +1,18 @@
+title: Experiment Creation
+description: 'Data required to create an experiment object'
```
aerlaut:
```suggestion
description: 'Data required to create a new experiment'
```

src/loaders/sqlClient.js
```diff
@@ -12,7 +12,7 @@ module.exports = async () => {
 
   const knexConfig = (await knexfile())[config.clusterEnv];
 
-  logger.log(`SQL endpoint at: ${knexConfig.connection.host}`);
+  logger.log(`sql endpoint at: ${knexConfig.connection.host}`);
```
aerlaut, line 15: why not uppercase? SQL is an abbreviation, so it should be all caps

src/loaders/express.js
```diff
@@ -110,19 +109,28 @@ module.exports = async (app) => {
...
-        AWSXRay.getSegment().addAnnotation('openApiValidationFailed', true);
-      },
-    },
-    operationHandlers: path.join(__dirname, '..', 'api'),
-  }));
+  ['v1', 'v2'].forEach((version) => {
+    // api and api.v2 could be renamed to api.v1 api.v2 once we focus on data model,
+    // for now leaving like this so we don't get merge conflicts
+    // with other changes that may come in on "api/"
+    const apiYamlFileName = version === 'v1' ? 'api' : `api.${version}`;
```
aerlaut: I think this variable is better called `apiVersion`. it doesn't store the filename. The filename is created on line 120.

src/api.v2/routes/health.js
aerlaut: Do we have other functions that make sense to be grouped under the same controller file?

aerlaut: If not then I think it's reasonable to have this route as is.

src/api.v2/helpers/modelGuts.js
```diff
@@ -0,0 +1,66 @@
+const sqlClient = require('../../sql/sqlClient');
+
+// The guts of a model that uses Knexjs to store and retrieve data from a
```
aerlaut: I understand that 'guts' is a Knex term, but is there a better name that is more common? The purpose of this module is to initialize an object with actions. How about `initializeModel` or `createModel`?

src/api.v2/model/userAccess.js
```diff
@@ -0,0 +1,19 @@
...
+];
+
+const guts = createGuts({
+  tableName,
+  selectableProps,
+});
+
+module.exports = {
+  ...guts,
+};
```
aerlaut: Are we spreading because we want a new instace of the object?

src/api.v2/helpers/modelGuts.js
```diff
@@ -0,0 +1,66 @@
...
+// composed on top of this set of common guts.
+//
+// The idea is that these are the most-used types of functions that most/all
+// "models" will want to have.
+
+// Copied from https://github.com/robmclarty/knex-express-project-sample/blob/main/server/helpers/model-guts.js
+module.exports = ({
+  tableName,
+  selectableProps = [],
+  timeout = 1000,
```
aerlaut: Is this in milliseconds? Is it enough time?

src/api.v2/controllers/experimentController.js
```diff
@@ -0,0 +1,32 @@
...
+  const { name, description } = body;
+
+  await experiment.create({ id: experimentId, name, description });
+
+  await Promise.all([
+    userAccess.create(
+      { user_id: user.sub, experiment_id: experimentId, access_role: AccessRole.OWNER },
+    ),
+    userAccess.create(
+      { user_id: config.adminArn, experiment_id: experimentId, access_role: AccessRole.OWNER },
```
aerlaut: I think `config.adminArn` is better named `config.adminSub` right?

src/api.v2/controllers/experimentController.js
```diff
@@ -0,0 +1,32 @@
...
+const config = require('../../config');
+
+const experiment = require('../model/experiment');
+const userAccess = require('../model/userAccess');
+
+const AccessRole = require('../../utils/enums/AccessRole');
+
+const { OK } = require('../../utils/responses');
+
+const createExperiment = async (req, res) => {
```
ivababukova, line 15: can you import the logger in here and log some messages that will be useful for us for debugging? Like "Creating experiment ${experimentId}" and "${experimentId} created successfully" and so on?

src/api.v2/helpers/modelGuts.js
```diff
@@ -0,0 +1,66 @@
+const sqlClient = require('../../sql/sqlClient');
```
ivababukova: this is so beautiful 😍 Goodbye, hacky dynamodb inserts, custom for each field and table!

src/api.v2/controllers/experimentController.js
```diff
@@ -0,0 +1,32 @@
...
+  await Promise.all([
+    userAccess.create(
+      { user_id: user.sub, experiment_id: experimentId, access_role: AccessRole.OWNER },
+    ),
+    userAccess.create(
+      { user_id: config.adminArn, experiment_id: experimentId, access_role: AccessRole.OWNER },
+    ),
+  ]);
+
+  res.json(OK());
```
ivababukova, line 36: what about handling the cases when create fails? Should we handle exceptions in here or in `modelGuts.js`?

src/api.v2/model/inviteAccess.js
```diff
@@ -0,0 +1,19 @@
...
+const tableName = 'invite_access';
+
+const selectableProps = [
+  'user_email',
+  'experiment_id',
+  'access_role',
+  'updated_at',
+];
+
+const guts = createGuts({
```
ivababukova: this `createGuts` function is actually not implemented in `modelGuts` or anywhere else. How does this work then?

src/api.v2/controllers/experimentController.js
ivababukova: why do we need to insert the same user twice? What is `user.sub` equal to?

src/loaders/sqlClient.js
cosa65: Since we ended up using lowercase for `sqlClient` and `sql` folders I added this change for coherency

src/loaders/express.js
cosa65: I could change it to `versionName` I think but not just `apiVersion` because it is a synonym of `version` which is a different variable

src/api.v2/helpers/modelGuts.js
cosa65: I don't think it is a knex term, it might just be the word guts.

It adds some basic queries, so maybe if you agree we could call it `basicModelFunctions`

src/api.v2/model/userAccess.js
cosa65: we are spreading to get all the properties of the object, not the object itself

src/api.v2/helpers/modelGuts.js
cosa65: Maybe we could make it longer, what number would you suggest? 3 or 4 seconds maybe?

cosa65: one of these simple sql queries taking longer than 2/3 seconds would be a little bit worrying though

src/api.v2/controllers/experimentController.js
cosa65: true, we coudl change, would it be ok @kafkasl?

cosa65: also I am wondering if it would make sense to store the adminArn somehwhere else (not statically in the code anyone can see). I know it shouldn't matter too much since it's not really useful on its own, but just in case.

src/api.v2/controllers/experimentController.js
cosa65: I think I would look into exceptions handling in a different ticket, we didn't have a clear exceptions handling policy in dynamodb in v1 so I think defining that now would take some extra work. Also on its own it doesn't help too much unless we check the different types of exceptions in the UI and display different errors for each kind of error.

src/api.v2/controllers/experimentController.js
cosa65: This is copying the behaviour in the v1 variant of `experimentCreation`, one user is the creator of the experiment, the other is the admin

cosa65: user.sub is the id assigned through congito to users

src/api.v2/routes/health.js
cosa65: > Do we have other functions that make sense to be grouped under the same controller file?

This one was just copied over from api v1, and it has only that endpoint (src/api/routes/health)

src/api.v2/helpers/modelGuts.js
cosa65: renamed it to `generateBasicModelFunctions`

src/api.v2/controllers/experimentController.js
aerlaut: I agree. I was thinkign of having some kind of config server. AWS has this functionality actually: https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html

tests/api.v2/routes/experiment.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+  afterEach(() => {
+    /**
+     * Most important since b'coz of caching, the mocked implementations sometimes does not reset
+     */
+    jest.resetModules();
+    jest.restoreAllMocks();
+  });
+
+  it('Creating a new experiment results in a successful response', async (done) => {
```
aerlaut, line 23: Maybe add tests for invalid responses? E.g. those with empty body, invalid body, invalid method

src/api.v2/controllers/experimentController.js
cosa65: we shoudl make a ticket for this maybe, or bring it up in the standup first

src/api.v2/helpers/modelGuts.js
aerlaut: If we have a good model, these queries should be fast, but we're also waiting for network latency, etc. right, so maybe 4 seconds?

cosa65: ok

aerlaut (approved):

PR #324, opened 2022-03-21
Patch and get endpoints experiment v2

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1803

https://ui-martinfosco-ui663-api324.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/663

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/specs/api.v2.yaml
```diff
@@ -76,21 +107,106 @@ paths:
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                oldPosition: 
+                  type: integer
+                newPosition:
+                  type: integer
```
ivababukova, line 175: you could explicitly add here "additionalProperties: false" to not allow passing of any parameters outside of the schema

src/api.v2/model/experiment.js
```diff
@@ -1,22 +1,69 @@
...
+
+  const result = await aggregateIntoJson(query, experimentFields, experimentExecutionFields, 'pipeline_type', 'pipelines', sql);
+
+  return result[0];
+};
+
+const updateSamplePosition = async (experimentId, oldPosition, newPosition) => {
+  const sql = sqlClient.get();
+
+  // Switches values between the item at newPosition and the one at oldPosition
```
ivababukova: is the idea that the initial order of the samples will be added to the experiment table when we call create samples? How will we define what should be the initial order? Will we sort them alphabetically based on names?

src/utils/authMiddlewares.js
```diff
@@ -210,16 +212,26 @@ const authenticationMiddlewareSocketIO = async (authHeader) => {
...
   }
 
-  const granted = await accessService.canAccessExperiment(userId,
-    experimentId,
-    resource,
-    method);
+  let granted;
+
+  if (isV2) {
+    granted = await userAccess.canAccessExperiment(userId, experimentId);
```
ivababukova: what about the more granular access permissions based on what resource and method the user wants to access? The ones that v1 version of `canAccessExperiment` uses and the `isRoleAuthorized` function in it?

tests/api.v2/model/experiment.test.js
```diff
@@ -14,7 +14,7 @@ describe('model/experiment', () => {
         'id',
         'name',
         'description',
-        'processing_config',
```
ivababukova, line 17: where did `processing_config` go?

tests/sql/helpers.test.js
```diff
@@ -0,0 +1,59 @@
...
+      'samples_order as samplesOrder',
+      'notify_by_email as notifyByEmail',
+      'created_at as createdAt',
+      'updated_at as updatedAt',
+    ]);
+  });
+});
+
+describe('aggregateIntoJson', () => {
+  it('Works correctly', async () => {
```
ivababukova, line 4: what other cases can we test for in here?

src/api.v2/controllers/experimentController.js
```diff
@@ -31,11 +44,47 @@ const createExperiment = async (req, res) => {
...
+
+  res.json(OK());
+};
+
+const patchExperiment = async (req, res) => {
+  const { params: { experimentId }, body } = req;
+
+  logger.log(`Updating experiment ${experimentId}`);
+
+  const snakeCasedKeysToPatch = _.mapKeys(body, (_value, key) => _.snakeCase(key));
```
ivababukova, line 40: this will probably be used in other places too. Why don't you move it in `helpers` next to `sqlToCamelCased`? Or do you think that it will only be used in here?

tests/sql/helpers.test.js
```diff
@@ -0,0 +1,59 @@
+const { sqlToCamelCased, aggregateIntoJson } = require('../../src/sql/helpers');
+
+describe('sqlToCamelCased', () => {
```
ivababukova: what about a test for converting from camel case to snake case?

tests/api.v2/routes/experiment.test.js
```diff
@@ -3,6 +3,7 @@ const request = require('supertest');
 const expressLoader = require('../../../src/loaders/express');
 
 jest.mock('../../../src/api.v2/controllers/experimentController');
+jest.mock('../../../src/utils/authMiddlewares');
```
ivababukova: Are you planning to write tests for `patchExperiment` and `updateSamplePosition`?

tests/api.v2/controllers/experimentController.test.js
```diff
@@ -4,8 +4,13 @@ const AccessRole = require('../../../src/utils/enums/AccessRole');
...
+  id: 'mockExperimentId',
+  name: 'mockExperimentName',
+  description: 'mockExperimentDescription',
+};
 jest.mock('../../../src/api.v2/model/experiment', () => ({
-  create: jest.fn(),
+  create: jest.fn(() => Promise.resolve([mockExperiment])),
 }));
 jest.mock('../../../src/api.v2/model/userAccess', () => ({
   create: jest.fn(),
```
ivababukova: Are you planning to write tests for patchExperiment and updateSamplePosition?

tests/api.v2/controllers/experimentController.test.js
```diff
@@ -4,8 +4,13 @@ const AccessRole = require('../../../src/utils/enums/AccessRole');
 const experimentModel = require('../../../src/api.v2/model/experiment');
 const userAccessModel = require('../../../src/api.v2/model/userAccess');
 
+const mockExperiment = {
```
ivababukova, line 5: in here, we'll need to add some samples_order?

src/api.v2/model/experiment.js
cosa65: Right now we don't support batch addition of samples on the api side, we perform one samples create call for each sample we want to upload. The samples are just appended to the end.

I think we should have that same behavior for v2 and perhaps in the future think about implementing a more intelligent samples creation behavior once we are done with the bulk of v2

src/utils/authMiddlewares.js
cosa65: You're right, added role checking

tests/api.v2/model/experiment.test.js
cosa65: I had removed it because the UI doesn't need the processing config in getExperiment (it isn't returned in the v1 implementation)

cosa65: the UI loads the basic experiment's data and sets processing with a loading flag so that the processing config is loaded.

I believe this is because the GET endpoint for a single experiment is used in `getExperimentInfo` in the UI, this actions is called on the server side of the UI and I guess it was made to avoid loading too much stuff from the server side, just the essentials instead.

src/api.v2/controllers/experimentController.js
cosa65: you're right, I'll move it

cosa65: I think it is not very sql-specific though, it is a very generic function compared to `sqlToCamelCased` so I think it fits more in utils

cosa65: perhaps we could leave it here and abstract it when we see how it can be reused in a new case when the new case shows up?

tests/sql/helpers.test.js
cosa65: depends on how [this](https://github.com/hms-dbmi-cellenics/api/pull/324#discussion_r832999964) is resolved

src/api.v2/model/experiment.js
ivababukova: makes sense

src/api.v2/controllers/experimentController.js
ivababukova: sure, we can leave it here for now

tests/api.v2/routes/experiment.test.js
cosa65: Added tests

tests/api.v2/controllers/experimentController.test.js
cosa65: Added tests

tests/sql/helpers.test.js
cosa65: It doesn't have any ifs or error handling so I didn't really see the need to add another test

package.json
```diff
@@ -9,7 +9,7 @@
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand --silent",
     "coverage": "jest --coverage --silent",
     "dev": "nodemon --max-old-space-size=8192 src/app.js",
-    "start": "node --max-old-space-size=8192 src/app.js",
+    "start": "node --unhandled-rejections=strict --max-old-space-size=8192 src/app.js",
```
aerlaut: What will adding this flag do?

src/api.v2/model/experiment.js
```diff
@@ -1,22 +1,91 @@
...
+  const sql = sqlClient.get();
+
+  // Sets samples_order as an array that has the sample in oldPosition moved to newPosition
+
+  const trx = await sql.transaction();
+
+  try {
+    const result = await trx(experimentTable).update({
+      samples_order: trx.raw(`(
+        SELECT jsonb_insert(samples_order - ${oldPosition}, '{${newPosition}}', samples_order -> ${oldPosition}, false)
```
aerlaut, line 68: Can you please explain what each of the parameter values do on this line?

src/sql/sqlClient.js
```diff
@@ -1,11 +1,22 @@
 const knex = require('knex');
+const { types } = require('pg');
+
+const TIMESTAMPZ = 1184;
+
+const setCustomParsers = () => {
+  // Default timestampz parser returns an object which breaks swagger's validation
+  // With this we can override that parser to return a string
+  types.setTypeParser(TIMESTAMPZ, (dateString) => dateString);
```
aerlaut, line 9: Does this also set the returned object's type to string?

src/utils/authMiddlewares.js
```diff
@@ -210,16 +212,33 @@ const authenticationMiddlewareSocketIO = async (authHeader) => {
  * and remove authByExperiment
  */
 const authorize = async (userId, resource, method, authResource, authByExperiment = true) => {
+  const isV2 = resource.startsWith('/v2/');
+
   let experimentId = authResource;
-  if (!authByExperiment) {
+
+  // authResource is always experimentId in V2
```
aerlaut: Because there is no project?

tests/api.v2/model/experiment.test.js
```diff
@@ -1,30 +1,121 @@
...
+
+    expect(mockTrx).toHaveBeenCalledWith('experiment');
+
+    expect(mockTrx.update).toHaveBeenCalledWith({ samples_order: 'resultOfSql.raw' });
+    expect(mockTrx.raw.mock.calls[0]).toMatchSnapshot();
+    expect(mockTrx.where).toHaveBeenCalledWith('id', 'mockExperimentId');
+    expect(mockTrx.returning).toHaveBeenCalledWith(['samples_order']);
+
+    expect(mockTrx.commit).toHaveBeenCalled();
+    expect(mockTrx.rollback).not.toHaveBeenCalled();
```
aerlaut, line 91: Is there an easy way to check the calls to these functions instead of doing them one by one?

package.json
cosa65: It transforms unhandled rejections from just showing up in the log to actually throwing the error that caused the rejection, I was debugging some problems with the validation yaml that weren't really apparent (the api still seemed to work) until I noticed some lines about unhandled rejections.

I thought it could be useful to keep this after I was done with debugging that.

Although I understand that perhaps it is better to do some more in depth testing on how adding this flag could affect production before actually introducing this, let me know what you think

src/api.v2/model/experiment.js
cosa65: in `jsonb_insert`?

It can work for both objects and arrays, but since in this case we are working with a jsonb array (`samples_order`) I'll explain it as an array.

It takes 3/4 parameters (last bool param is optional) and returns a jsonb array with a new element inserted.

1st param: original array
2nd param: (format: `'{5}'`), the index at which you want to insert the value.
3rd param: The value that we want to insert.
4th param: "Insert after", determines whether the new value is inserted before or after the element that already exists at `newPosition`.

For 1st param: `array - index` returns the result pf removing the value at `index` from `array`, so we are actually moving the value at `oldPosition` instead of just inserting another copy of it

cosa65: None of this modifies anything in the actual `samples_order`, what actually does the modificationg is wrapping all of this in a SQL `UPDATE`

src/sql/sqlClient.js
cosa65: dateString is a string already

cosa65: the problem is that without doing this we don't get `dateString`, we get a wrapper that transforms to a string when we stringify but `typeof` returns object. 

This shouldn't actually be a problem except for the fact that the openapi validation checks the type before doing the stringify I believe, so it sees it isn't a string and complains.

src/utils/authMiddlewares.js
cosa65: Yep

aerlaut: Can you add that in the comments?

package.json
aerlaut: Does "actually throwing the error" means causing the API process to stop?

src/api.v2/model/experiment.js
aerlaut: Oh I see... I was confused with `array - index`. I thought it was subtracting 2 integers.

package.json
cosa65: I think it might be safer to just remove it since I only actually used it for debugging, so there's no real reason for adding it into master

tests/api.v2/model/experiment.test.js
cosa65: Talked over in slack, we looked at some options and decided to leave it to another ticket

src/api.v2/controllers/experimentController.js
```diff
@@ -1,41 +1,101 @@
...
+  }
+
+  await experiment.updateSamplePosition(experimentId, oldPosition, newPosition);
+
+  logger.log(`Finished reordering samples in ${experimentId}`);
 
   res.json(OK());
 };
 
 module.exports = {
```
kafkasl, line 71: why export each function instead of exporting the whole experimentController with the functions as methods?

src/sql/helpers.js
```diff
@@ -0,0 +1,95 @@
+const _ = require('lodash');
+
+// Adds commands in a select to transform fields that are in snake_case format to camelCase
+const sqlToCamelCased = (snakeCasedFields) => snakeCasedFields.map((snakecased) => `${snakecased} as ${_.camelCase(snakecased)}`);
```
kafkasl: if it transforms snake_case to cameCase, I think it's better to name it `snakeToCamelCase` instead of `sql`

src/api.v2/controllers/experimentController.js
```diff
@@ -1,41 +1,101 @@
...
+const { NotFoundError } = require('../../utils/responses');
 
 const logger = getLogger('[ExperimentController] - ');
 
+const getExperiment = async (req, res) => {
+  const { params: { experimentId } } = req;
+
+  const data = await experiment.getExperimentData(experimentId);
+
+  if (_.isNil(data)) {
```
kafkasl: I think this exception shoulud be thrown by `experiment.getExperimentData(experimentId)`. If there is no experiment and you attempt to get it, throw an exception (which won't be handled here)

src/api.v2/controllers/experimentController.js
```diff
@@ -1,41 +1,101 @@
...
   const { params: { experimentId }, user, body } = req;
 
   const { name, description } = body;
 
   logger.log('Creating experiment');
 
-  await experiment.create({ id: experimentId, name, description });
+  const experimentCreateResult = await experiment.create({ id: experimentId, name, description });
+
+  if (experimentCreateResult.length === 0) {
```
kafkasl: again I think this error handling should be in the model method. I would accept handling it here if we could "do something" about it, to fix or handle the error but we are just throwing the error.

src/api.v2/controllers/experimentController.js
```diff
@@ -1,41 +1,101 @@
...
 
   logger.log('Setting up access permissions for experiment');
-  await Promise.all([
+  const userAccessCreateResults = await Promise.all([
     userAccess.create(
       { user_id: user.sub, experiment_id: experimentId, access_role: AccessRole.OWNER },
     ),
     userAccess.create(
-      { user_id: config.adminSub, experiment_id: experimentId, access_role: AccessRole.OWNER },
+      { user_id: config.adminSub, experiment_id: experimentId, access_role: AccessRole.ADMIN },
```
kafkasl: currently the `ADMIN` permissions are created AFTER the `OWNER` so that experiments created with the admin account have effectively no owner. Bioinformaticians use that to make it easier to then assign a new owner (without having to remove the original one). What will happen with these promises if the experiment & user are the same? I think doing them concurrently is not a good idea because the end state won't be deterministic.

src/api.v2/controllers/experimentController.js
```diff
@@ -1,41 +1,101 @@
...
+
+  res.json(OK());
+};
+
+const patchExperiment = async (req, res) => {
+  const { params: { experimentId }, body } = req;
+
+  logger.log(`Updating experiment ${experimentId}`);
+
+  const snakeCasedKeysToPatch = _.mapKeys(body, (_value, key) => _.snakeCase(key));
```
kafkasl, line 40: I am not sure it's a good idea to do all these transformations implicitly as opposed to explicitly constructing the update object. Some cons:
* Doing it like this makes it much harder to read or understand what's actually being patched. 
* If any API schema is not strict and perfectly matches the SQL schema, this will make the patch crash.

src/api.v2/model/experiment.js
```diff
@@ -1,22 +1,91 @@
...
+      .leftJoin(experimentExecutionTable, `${experimentTable}.id`, `${experimentExecutionTable}.experiment_id`)
+      .where('id', experimentId)
+      .as('experiment_with_exec');
+  }
+
+  const experimentExecutionFields = [
+    'params_hash', 'state_machine_arn', 'execution_arn',
+  ];
+
+  const result = await aggregateIntoJson(query, experimentFields, experimentExecutionFields, 'pipeline_type', 'pipelines', sql);
```
kafkasl: I'm a bit concerned that this `aggregateIntoJson` is complex enough that other people will avoid using it. It's also kind of hard to read. I've read the documentation and it's still kind of confusing, specially considering that in this case what it needs to do is pretty simple. Do you think this complexity is worth it? We just have 2 extra fields to add. You could only query for the experiment info in the first query and then do
```
experimentInfo = await ....

executions = await this.select('*')
      .from(experimentExecutionTable)
      .where('experiment_id', experimentId)

executions.forEach((exc) => {
    experimentInfo[exc.pipeline_type] = exc
}

```

src/api.v2/model/experiment.js
kafkasl, line 68: I think after this PR it'd be good to organize a short demo on basic stuff that we need to know for SQL. Like the `jsonb`, minimal migrations functionality, etc... I can help with it.

src/api.v2/model/userAccess.js
```diff
@@ -10,10 +13,27 @@ const selectableProps = [
...
+    .select()
+    .from(userAccessTable)
+    .where({ experiment_id: experimentId, user_id: userId });
+
+  // If there is no entry for this user and role, then user definitely doesn't have access
+  if (results.length === 0) {
+    return false;
+  }
+
+  const { access_role: role } = results[0];
```
kafkasl: you can  use this `first` when trying to get a single value instead of `[0]` https://github.com/knex/knex/issues/1355#issuecomment-211419629

src/specs/api.v2.yaml
```diff
@@ -76,21 +107,110 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPError'
+        '401':
+          description: The request lacks authentication credentials.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPError'
+    patch:
```
kafkasl, line 123: as I mentioned above, you are using "implicit" schema transformations for the patch/put operations, meaning that you just change the case of all the fields to update, and run the query. If we end up doing that, then the chemas here should be set to strict or those queries will crash if the UI sends unexpected or extra fields (which is fairly probable)

kafkasl: I've seen you've done that in most of the cases with the `additionalProperties: false`. Do you know if it's possible to make it the default for all or something like that?

src/sql/helpers.js
```diff
@@ -0,0 +1,95 @@
...
+ * ]
+ * For this example:
+ * - rootFields: ['id', 'name'],
+ * - nestedFields: ['pipeline_arn', 'hash]
+ * - aggregationColumnKey: 'pipeline_type'
+ * - aggregationJsonKey: 'pipelines'
+ *
+ * Note it also camelcases all the keys returned
+ */
+const aggregateIntoJson = async (
```
kafkasl: I commented above that so far I think this is too complex for the use case I've seen. I think it's easier to just add the desired fields with a less "generic" method.

src/utils/authMiddlewares.js
```diff
@@ -210,16 +212,33 @@ const authenticationMiddlewareSocketIO = async (authHeader) => {
  * and remove authByExperiment
  */
 const authorize = async (userId, resource, method, authResource, authByExperiment = true) => {
+  const isV2 = resource.startsWith('/v2/');
```
kafkasl: I thought these files would be duplicated too instead of having to modify one to support both.

src/api.v2/controllers/experimentController.js
cosa65: you mean wrapping everything in a class or something and exporting that? why would it be better?

src/sql/helpers.js
cosa65: It isn't only transforming snake_case, is creates commands so that sql transforms it:

"samples_order" => "samples_order as samplesOrder"

src/api.v2/controllers/experimentController.js
cosa65: with the way it is implemented right now, it will fail if you try to add two different roles for the same user and experiment, which I think makes sense.

I don't think doing them concurrently will change things except on this case you mentioned.

As a counter proposal, I think maybe I could add handling specifically for the case where `config.adminSub` === `user.sub` (only create the admin as owner, right?)

src/api.v2/model/experiment.js
cosa65: I agree, we talked about this with iva as well

cosa65: in particular explaining how to run the migrations locally and manage the local infrastructure in general (migrating dynamo data into the db as well)

src/specs/api.v2.yaml
cosa65: I had looked at making it the default a long time ago, I didn't find any way to do that

src/utils/authMiddlewares.js
cosa65: so you are proposing moving this into each of the api's folders and making one for each version?

src/sql/helpers.js
cosa65: discussing with agi we thought it could be useful in other use cases to facilitate avoiding doing the processing in the api.

src/api.v2/controllers/experimentController.js
cosa65: let me know what you think about the idea

src/api.v2/controllers/experimentController.js
cosa65: I could throw a generic error in the `generateBasicModelFunctions` file, since this is one of the generic functions, do you think I should do that?

src/api.v2/controllers/experimentController.js
kafkasl: when you say it will fail when adding two roles, you mean using this Promise.all or in general? because changing permissions should always be an overwriting operations, if you set a different role for a user that already has one, it should overwrite not fail. I'm ok with adding specific handling for the admin.

src/api.v2/controllers/experimentController.js
kafkasl: yes, reasons that come to mind are:
1. we were already doing that with all the experiment, projects, etc... services
2. naming conventions demand the file to be named as the exports (in this case it's not possible because there are many different functions.
3. when using this functions from the outside I think it's easier to "understand" them if they are grouped under the same import like `experimentController.getExperiment` & `experimentController.patchExperiment`. 

what would be the cons of doing it this way, and the pros of doing it as you did?

src/sql/helpers.js
kafkasl: I think we shouldn't be adding this extra logic around the code. afaict we can use something simpler like this https://vincit.github.io/objection.js/recipes/snake-case-to-camel-case-conversion.html

src/utils/authMiddlewares.js
kafkasl: I just hadn't noticed this. I feel that it's gonna become very messy if we migrate the whole code base like this because there are a lot of things in utils. What do you think?

src/sql/helpers.js
kafkasl: I honestly would be hard-pressed to find a situation in which I would prefer to use this complex thing over something like 
```
experimentInfo = await ....

executions = await this.select('*')
      .from(experimentExecutionTable)
      .where('experiment_id', experimentId)

executions.forEach((exc) => {
    experimentInfo[exc.pipeline_type] = exc
}

```

it hides structure and implementation details (making it harder to figure out what's happening) and you only have to look at how much comments you needed to describe what's happening instead of just writing the simple code I posted. Moreover, if we ever change the nesting structure of any of the many calls using this (if as you say, we end up using it) it will either generalize poorly or become hard to debug. In my opinion this violates clearly the KISS principle but it's up to you.

src/utils/authMiddlewares.js
cosa65: Ok, I'll move it

src/api.v2/controllers/experimentController.js
cosa65: > if you set a different role for a user that already has one, it should overwrite not fail.

Sure, but this isn't the case when we are in this function ,right? Doing an insert (not overwriting) makes sense here I think.

Edit:
> when you say it will fail when adding two roles, you mean using this Promise.all or in general?

I mean in both places, because the pair experimentid and userid needs to be unique.

cosa65: > I'm ok with adding specific handling for the admin.

Ok, I'll do that then.

src/sql/helpers.js
cosa65: It says there that `objection.js` is an ORM, it seems a little overkill to add this package only to use the mapper.

src/api.v2/controllers/experimentController.js
cosa65: 1. In the experiment, projects, etc services we were using classes without any need, no parameters were taken and all they did was hold the functions. There were other things like the naming (with underscores instead of camelcase) which we didn't like either and were changed.
2. I don't understand this, they are many functions wrapped in an object, we are exporting the object
3. We can still do that, when you do `require()` you can just do 
`const experimentController = require('../controllers/experimentController');`
instead of 
```
const {
  createExperiment, getExperiment, patchExperiment, updateSamplePosition,
} = require('../controllers/experimentController');
```

The cons is that I don't see why we would use a class when it doesn't add any value.
The pros is that it's the way we have been handling export and import for files that hold many objects in most places in the UI and the api (the only exception I can think of are the `*-services`).

src/sql/helpers.js
cosa65: oh I see, we could do the camelcase conversion in `postProcessResponse`, good catch!

src/sql/helpers.js
cosa65: resolved in a slack call

src/api.v2/middlewares/authMiddlewares.js
```diff
@@ -0,0 +1,72 @@
...
+ * Calls `authorize()` internally.
+ */
+const expressAuthorizationMiddleware = async (req, res, next) => {
+  if (!req.user) {
+    next(new UnauthenticatedError('The request does not contain an authentication token.'));
+    return;
+  }
+
+  const authByExperiment = !!req.params.experimentId;
+  const authResource = authByExperiment ? req.params.experimentId : req.params.projectUuid;
```
kafkasl: there are only experiment Ids now, no need for this

tests/api.v2/model/userAccess.test.js
```diff
@@ -1,27 +1,171 @@
...
+
+const mockCreate = jest.fn();
+jest.mock('../../../src/api.v2/helpers/generateBasicModelFunctions', () => jest.fn(() => (
+  {
+    hasFakeBasicModelFunctions: true,
+    create: mockCreate,
+  }
+)));
+
+// const mockIsRoleAuthorized = jest.fn();
```
kafkasl: delete

kafkasl (approved):

PR #327, opened 2022-03-28
Load all experiments api v2

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1803

https://ui-martinfosco-ui673-api327.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/673

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api.v2/model/experiment.js
```diff
@@ -93,6 +126,7 @@ const updateSamplePosition = async (experimentId, oldPosition, newPosition) => {
 };
 
 module.exports = {
+  getAllExperiments,
   getExperimentData,
```
ivababukova, line 155: I see that for `getExperimentData` and `getAllExperiments` we don't use transactions. Is this because these are not writing data to the database and there is no need to have them atomic?

src/api/route-services/access/index.js
```diff
@@ -79,7 +79,7 @@ class AccessService {
 
     const dynamodb = createDynamoDbInstance();
 
-    await dynamodb.deleteItem(dynamoParams).send();
+    await dynamodb.deleteItem(dynamoParams).promise();
```
ivababukova, line 112: what did `send()` do?

src/sql/helpers.js
```diff
@@ -64,26 +64,44 @@ const jsonbObjectAgg = (aggregationColumnKey, nestedFields, aggregationJsonKey,
...
-const aggregateIntoJson = (
+const collapseKeysIntoObject = (
   originalQuery,
   rootFields,
-  nestedFields,
-  aggregationColumnKey,
-  aggregationJsonKey,
+  fieldsToCollapse,
+  collapseKey,
+  collapseKeyNewName,
```
ivababukova: rename `collapseKeyNewName` to `collapsedKeyNewName` to match the name in `collapseKeyIntoArray` function

src/sql/helpers.js
```diff
@@ -1,4 +1,4 @@
-const jsonbObjectAgg = (aggregationColumnKey, nestedFields, aggregationJsonKey, sql) => {
+const createCollapsedObject = (collapseKey, nestedFields, collapseKeyNewName, sql) => {
```
ivababukova: rename collapseKeyNewName to collapsedKeyNewName to match the name in collapseKeyIntoArray function

src/sql/helpers.js
```diff
@@ -64,26 +64,44 @@ const jsonbObjectAgg = (aggregationColumnKey, nestedFields, aggregationJsonKey,
...
       ...rootFields,
-      jsonbObjectAgg(aggregationColumnKey, nestedFields, aggregationJsonKey, sql),
+      createCollapsedObject(collapseKey, fieldsToCollapse, collapseKeyNewName, sql),
     ])
     .from(originalQuery)
     .groupBy(rootFields)
 );
 
-module.exports = { aggregateIntoJson };
+const collapseKeyIntoArray = (
```
ivababukova, line 1: rename `collapseKeyIntoArray` to `collapseKeysIntoArray` to match the naming of `collapseKeysIntoObject`

src/sql/helpers.js
```diff
@@ -64,26 +64,44 @@ const jsonbObjectAgg = (aggregationColumnKey, nestedFields, aggregationJsonKey,
...
- * - aggregationColumnKey: 'pipeline_type'
- * - aggregationJsonKey: 'pipelines'
+ * - collapseKey: 'pipeline_type'
+ * - collapseKeyNewName: 'pipelines'
  *
  * Note it also camelcases all the keys returned
  */
-const aggregateIntoJson = (
+const collapseKeysIntoObject = (
   originalQuery,
```
ivababukova: why don't you rename `originalQuery` and `mainQuery` names used in here and in `model/experiment.js` to all be called `query` or `mainQuery`? If they match names, the code will be clearers that they are the same thing. `query` can be a good name, because they are all essentially the query. Just a suggestion

src/sql/helpers.js
```diff
@@ -10,32 +10,32 @@ const jsonbObjectAgg = (aggregationColumnKey, nestedFields, aggregationJsonKey,
...
  * @param {*} originalQuery Query that fetches the sql data
  * @param {*} rootFields Fields that shouldn't be aggregated
- * @param {*} nestedFields Fields that vary across each aggregationColumnName,
+ * @param {*} fieldsToCollapse Fields that vary across each aggregationColumnName,
  * this is what we want to aggregate into one object for each aggregationColumnName
- * @param {*} aggregationColumnKey The column by which we want to perform the aggregation,
+ * @param {*} collapseKey The column by which we want to perform the aggregation,
  * the object will contain one key for each aggregationColumnName value
- * @param {*} aggregationJsonKey The key that we want to be added holding all the aggregated data
+ * @param {*} collapseKeyNewName The key that we want to be added holding all the aggregated data
```
ivababukova: These `collapseKeysIntoObject` and `collapseKeyIntoArray` methods seem really complex and are hard to read. I can't come up with a way to make them easier to read, I think that the problem is the generalisation. Do we really need to have "magic" methods like this that can handle all sorts of fields? Does the benefit of reducing code outweigh the addition of extra complexity?

src/specs/models/experiment-bodies/ProcessingConfig.v2.yaml
```diff
@@ -0,0 +1,28 @@
...
+    $ref: ../processing-config-bodies/ProcessingConfigNumGenesVsNumUMIs.v1.yaml
+  doubletScores:
+    $ref: ../processing-config-bodies/ProcessingConfigDoubletScores.v1.yaml
+  dataIntegration:
+    $ref: ../processing-config-bodies/ProcessingConfigDataIntegration.v1.yaml
+  configureEmbedding:
+    $ref: ../processing-config-bodies/ProcessingConfigConfigureEmbedding.v1.yaml
+  meta:
+    $ref: ../processing-config-bodies/ProcessingConfigMeta.v1.yaml
+required:
```
ivababukova, line 21: can you add `additionalProperties: false`?

ivababukova: to avoid the ui passing fields that we don't support

src/sql/helpers.js
cosa65: In this case no because we are collapsing only one key, while in the object we are collapsing many different keys.

src/specs/models/experiment-bodies/ProcessingConfig.v2.yaml
cosa65: Before doing this we will need to clean up the processingconfigs in dynamo (or after migrating) because they have a lot of stuff.

Perhaps I should perform that cleanup in the dynamoToSql migration script?

ivababukova: as discussed on a call, we can make the schema stricter as part of separate piece of work

src/api/route-services/access/index.js
cosa65: As far as I could see the same thing, only that after doing `npm install` to fix the `package-lock` problems `aws-sdk-mock` didn't manage commands sent with `send` correctly while `.promise()` it does (`promise()` is the one we use the most).

[Looking in the docs](https://aws.amazon.com/blogs/developer/support-for-promises-in-the-sdk/#:~:text=Making%20Requests%20by%20Using%20Promises) it doesn't really say if there is any difference when it mentions both.

> You could call send() on the request to make the service call.

> The promise() method immediately starts the service call and returns a promise.

They don't mention that send() also returns a promise, so honestly based on this it is basically doing the same thing.

src/api.v2/model/experiment.js
cosa65: it's only one query for either of these so there's no need to make them explicitly atomic, they already are.

src/api/route-services/access/index.js
ivababukova: sounds good, I was surprised because I didn't know about the existence of send

ivababukova (approved):

PR #330, opened 2022-04-07
Sample create delete api v2

cosa65:


src/api.v2/model/Experiment.js
```diff
@@ -0,0 +1,165 @@
...
+
+const getLogger = require('../../utils/getLogger');
+
+const logger = getLogger('[ExperimentModel] - ');
+
+const experimentTable = 'experiment';
+const experimentExecutionTable = 'experiment_execution';
+const userAccessTable = 'user_access';
+const metadataTrackTable = 'metadata_track';
+
```
ivababukova: We can move all these files to config to avoid duplicate definition of table names

src/specs/api.v2.yaml
```diff
@@ -188,6 +188,72 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPSuccess'
+        '400':
+          description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPError'
+        '401':
```
ivababukova, line 263: If I am not mistaken, the API returns 403 for non-authorized errors, not 401. Can you double-check?

tests/api.v2/controllers/sampleController.test.js
```diff
@@ -0,0 +1,59 @@
...
+const mockSampleId = 'mockSampleId';
+const mockSampleTechnology = '10x';
+const mockSampleName = 'mockSampleName';
+
+describe('sampleController', () => {
+  beforeEach(async () => {
+    jest.clearAllMocks();
+  });
+
+  it('createSample works correctly', async () => {
```
ivababukova, line 37: can you add a test for when one of the items under the createSample transaction fail, to test that the whole transaction is rollbacked and nothing gets created in this case?

ivababukova: Also, can you think of any other possible test cases for this createSample function? What else can go wrong?

tests/api.v2/model/BasicModel.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+const mockRowId = 'mockId';
+const mockRowName = 'mockName';
+
+describe('model/BasicModel', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('create works correctly', async () => {
```
ivababukova, line 21: I don't see any tests for failure modes. Is there something we can test for?

tests/api.v2/model/Experiment.test.js
```diff
@@ -172,4 +162,26 @@ describe('model/experiment', () => {
...
+    mockSqlClient.raw.mockImplementationOnce(() => 'RawSqlCommand');
+
+    await new Experiment().addSample(mockExperimentId, mockSampleId);
+
+    expect(mockSqlClient.update).toHaveBeenCalledWith({ samples_order: 'RawSqlCommand' });
+    expect(mockSqlClient.raw).toHaveBeenCalledWith('samples_order || \'["mockSampleId"]\'::jsonb');
+    expect(mockSqlClient.where).toHaveBeenCalledWith('id', 'mockExperimentId');
+  });
+
+  it('deleteSample works correctly', async () => {
```
ivababukova, line 177: can we test for deleteSample and addSample failing? What other edge cases can we have?

ivababukova (approved):
I tested in staging, it is working well in v1. Didn't test v2 in staging. Feel free to merge after figuring out if you want to address my comments about the unit tests or if it is ok as they are.

tests/api.v2/model/BasicModel.test.js
cosa65: I don't really think there are any unit tests to add there since all `BasicModel` does is do part of a query

src/specs/api.v2.yaml
cosa65: 401 is lacking authentication credentials, 403 is forbidden, so credentials exist but not permission

cosa65: I'll add 403 as well

PR #340, opened 2022-04-26
[BIOMAGE-1870] - Implement loadRoles endpoint

aerlaut:
# Description
Implement API endpoint and controllers for `loadRoles` functionality

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1870

https://ui-agi-paltry-fish.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/696

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


__mocks__/@kubernetes/client-node.js
```diff
@@ -22,12 +22,9 @@ const mockApi = {
   })),
 };
 
-k8s.KubeConfig.mockImplementation(() => {
-  console.debug('mocking the constructor');
```
aerlaut, line 26: Just removing this `console.log`

tests/api.v2/routes/access.test.js
```diff
@@ -0,0 +1,45 @@
...
+    request(app)
+      .get('/v2/access/nonExistentExperimentId')
+      .expect(404)
+      .end((err) => {
+        if (err) {
+          return done(err);
+        }
+        return done();
+      });
+  });
```
cosa65: I don't think this test is useful. Isn't it just testing the mock? I feel like the only actual functionality that is being tested here is that the request reaches the controller (which is something that is alreayd tested by the previous test too).

src/api.v2/model/UserAccess.js
```diff
@@ -27,6 +28,38 @@ class UserAccess extends BasicModel {
...
+
+    const experimentUsers = cognitoUserData.map((userInfo, idx) => {
+      const email = userInfo.find((attr) => attr.Name === 'email').Value;
+      const name = userInfo.find((attr) => attr.Name === 'name').Value;
+      const { accessRole } = filteredAccess[idx];
+
+      return {
+        name, email, role: accessRole,
+      };
+    });
```
cosa65: What do you think of perhaps just making the model return the sql response.

Then make the controller call a function that does all this part that interacts with aws cognito.

We could store this cognito function under a `cognito` folder inside `helpers` (that I think will be renamed to services probably)

cosa65: That way model remains a sql-only thing

src/api.v2/routes/access.js
```diff
@@ -0,0 +1,13 @@
+const {
+  getExperimentUsers,
+} = require('../controllers/accessController');
+
+const { expressAuthenticationOnlyMiddleware, expressAuthorizationMiddleware } = require('../middlewares/authMiddlewares');
+
+module.exports = {
+  'access#getExperimentUsers': [
+    expressAuthenticationOnlyMiddleware,
+    expressAuthorizationMiddleware,
```
cosa65: I believe the `expressAuthorizationMiddleware` already includes authentication

aerlaut: oh right!

tests/api.v2/routes/access.test.js
aerlaut: This test tests that the route will pass on an error properly (the part `.catch(next)`) in the route.

src/api.v2/model/UserAccess.js
aerlaut: This depends on how we want to separate the concerns. From what I read, in MVC frameworks, people usually advocate for a ["thin controller, fat model"](https://stackoverflow.com/questions/14044681/fat-models-and-skinny-controllers-sounds-like-creating-god-models). Arguably, the controller's work in an MVC framework does not only clean input and receive data from a model, but also create the view, which our controller is not handling and make our controller feels very "thin".

I feel that anything business-logic or data-related should be placed in the model, so that we have only one place where data is calculated and processed. Doing the Cognito fetches in the controller would spread data processing in two layers.

aerlaut: But this depends on how we want to structure the API.

tests/api.v2/routes/access.test.js
cosa65: Makes sense. 

Maybe then do `it("returns errors triggered by the controller")` so it's more general, becuase if not it feels like we need to also add more tests for different kinds. of errors, when all we care about is that any error is passed on next.

Also I think the pattern of passing different ids to define what kind of error to trigger is not ideal, I think it's better to just mock the controller function to trigger an error inside this test in particular. It's more concrete about what's happening

src/api.v2/model/UserAccess.js
cosa65: discussed over a call

tests/api.v2/routes/access.test.js
aerlaut: moved

cosa65 (approved):

PR #343, opened 2022-04-28
[BIOMAGE-1836] Implement metadata tracks-specific endpoints

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1836

https://ui-martinfosco-ui699-api343.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/699

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/api.v2/controllers/metadataTrackController.js
+++ b/src/api.v2/controllers/metadataTrackController.js
@@ -31,14 +31,34 @@ const patchMetadataTrack = async (req, res) => {
     .update({ experiment_id: experimentId, key: oldKey }, { key });
 
   if (result.length === 0) {
-    throw new NotFoundError(`Metadata track ${oldKey} doesn't exist`);
+    throw new NotFoundError(`Metadata track ${oldKey} not found`);
   }
 
   logger.log(`Finished patching metadata track ${oldKey} in experiment ${experimentId}, changed to ${key}`);
   res.json(OK());
 };
 
-const patchSampleInMetadataTrackValue = async (req, res) => {
+const deleteMetadataTrack = async (req, res) => {
+  const {
+    params: { experimentId, metadataTrackKey },
+  } = req;
+
+  logger.log(`Creating metadata track ${metadataTrackKey} in experiment ${experimentId}`);
+
+  const result = await new MetadataTrack().delete(
+    { experiment_id: experimentId, key: metadataTrackKey },
+  );
+
+  if (result.length === 0) {
+    throw new NotFoundError(`Metadata track ${metadataTrackKey} not found`);
+  }
+
+  logger.log(`Finished creating metadata track ${metadataTrackKey} in experiment ${experimentId}`);
+
+  res.json(OK());
+};
+
+const patchValueForSample = async (req, res) => {
   const {
     params: { experimentId, sampleId, metadataTrackKey },
     body: { value },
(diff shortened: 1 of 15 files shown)


src/api.v2/controllers/sampleController.js
```diff
@@ -52,7 +52,7 @@ const deleteSample = async (req, res) => {
   logger.log(`Deleting sample ${sampleId} from experiment ${experimentId}`);
 
   await sqlClient.get().transaction(async (trx) => {
-    await new Sample(trx).destroy(sampleId);
+    await new Sample(trx).delete(sampleId);
```
ivababukova: Will the metadata track entries, if any, be deleted when we call `Sample.delete` in line 55?

src/api.v2/controllers/metadataTrackController.js
```diff
@@ -0,0 +1,76 @@
...
+const patchMetadataTrack = async (req, res) => {
+  const {
+    params: { experimentId, metadataTrackKey: oldKey },
+    body: { key },
+  } = req;
+
+  logger.log(`Patching metadata track ${oldKey} in experiment ${experimentId}`);
+
+  const result = await new MetadataTrack()
+    .update({ experiment_id: experimentId, key: oldKey }, { key });
```
ivababukova, line 31: I see that `patchMetadataTrack` updates the `oldKey` to `newKey` and from what I understand by reading the rest of the code, the key is not equivalent to the value of metadata (for example, the initial and default value of `N.A.` is not key). Then in `patchMetadataTrack`, why do we want to update the key? Shouldn't we be updating the value of the metadata track instead? Why would you ever want to change the key of the metadata track?

ivababukova: Or maybe `metadataTrackKey` is the title of the track column, for example "my awesome_track" in this screenshot below?

Screenshot 2022-04-29 at 09 23 20

ivababukova: yes, this must be it

ivababukova: I wonder whether I am the only one that gets confused by metadata track value (the title of the metadata column) vs the metadata track value per sample?

src/api.v2/model/MetadataTrack.js
```diff
@@ -3,21 +3,49 @@ const BasicModel = require('./BasicModel');
...
+      if (sampleIds.length === 0) {
+        return;
+      }
+
+      const [{ id: metadataTrackId }] = response;
+
+      const valuesToInsert = sampleIds.map(({ id: sampleId }) => ({
+        metadata_track_id: metadataTrackId,
+        sample_id: sampleId,
+        value: 'N.A.',
```
ivababukova: you can create a constant for default metadata track sample value and use it in line 42 and in line 61. That way if we wanted to change the default value, we wouldn't need to remember that it needs to be changed in two places.
I was briefly considering moving lines 39-42 and lines 58-62 to a common function because they are pretty much the same, but I don't think it is really worth it.

src/api.v2/controllers/metadataTrackController.js
cosa65: I call the title `metadataTrackKey` or just `key` (not value), then `N.A.`cells I call the `value` (for a sample in a metadata track)

src/api.v2/model/MetadataTrack.js
cosa65: I could also set the default in schema actually so we don't need to have it in the api

src/api.v2/controllers/sampleController.js
cosa65: No, we tlaked about this with agi yesterday actually, you can test the current behavior in apj v1, if you remove all samples and add a new one after the metadata tracks will still be there.

(The `sample_in_metadata_track_map` rows will be auto deleted).

src/api.v2/model/MetadataTrack.js
cosa65: Done

src/api.v2/controllers/sampleController.js
ivababukova: I didn't know this about v1. How do `sample_in_metadata_track_map` get auto deleted? Is there a trigger?

src/api.v2/controllers/metadataTrackController.js
ivababukova: Sounds good, it just took me some time to figure out. One key with one or more values.

ivababukova (approved):

src/api.v2/controllers/sampleController.js
cosa65: it's the result of th eforeign keys set up.
If you write ON DELETETE CASCADE you're saying that when the foreign key a row is referencing is deleted then its own action should be to also be deleted.

PR #350, opened 2022-05-09
added kubernetes.v2 endpoint

kafkasl:
# Description
Added v2 endpoint for kubernetes events.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1881

N/A

https://github.com/hms-dbmi-cellenics/iac/pull/383

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api.v2/routes/kubernetes.js
```diff
@@ -0,0 +1,33 @@
...
+    logger.log('received kubernetes event');
+    try {
+      const {
+        reason, message, type, involvedObject: { name, namespace },
+      } = req.body;
+      logger.log(`[${reason}] received ${type} kubernetes event: ${message} ${name} in ${namespace}`);
+
+      // remove only pods in your namespace and due to backoff errors
+      if ((namespace.match('^pipeline-.*') || namespace.match('^worker-.*'))
+       && reason === 'BackOff' && type !== 'Normal' && message.includes('restarting')) {
```
StefanBabukov, line 22: Why does the condition need to be reason === BackOff and the message to include "restarting" is it not enough to delete the ones that are in BackOff?

tests/api.v2/routes/kubernetes.test.js
```diff
@@ -0,0 +1,107 @@
...
+      Logging: 'LoggingDisabled: LOGGING_CONFIGMAP_NOT_FOUND',
+    },
+  },
+};
+
+const express = require('express');
+const request = require('supertest');
+const expressLoader = require('../../../src/loaders/express');
+
+describe('tests for experiment route', () => {
```
StefanBabukov: kubernetes

tests/api.v2/routes/kubernetes.test.js
```diff
@@ -0,0 +1,107 @@
...
+    const mockApp = await expressLoader(express());
+    app = mockApp.app;
+  });
+
+  afterEach(() => {
+    /**
+     * Most important since b'coz of caching, the mocked implementations sometimes does not reset
+     */
+    jest.resetModules();
+    jest.restoreAllMocks();
```
StefanBabukov: I think you dont need these to run the test?

tests/api.v2/routes/kubernetes.test.js
kafkasl: lol I copy pasted it from the old test haha, changed

src/api.v2/routes/kubernetes.js
kafkasl: it turns out the `BackOff` is an umbrella term for many different issues. For example, when you are trying to download an image from dockerhub and you hit the rate limit, you'll get a backoff event (meaning that you are backing off and waiting before you retry the download again). After investigating this is the best way I managed to filter the restarts (documentation on kubernetes events is pretty bad and patchy, even checking the source code).

tests/api.v2/routes/kubernetes.test.js
kafkasl: removed

src/api.v2/routes/kubernetes.js
```diff
@@ -0,0 +1,33 @@
...
+  'kubernetes#event': async (req, res, next) => {
+    logger.log('received kubernetes event');
+    try {
+      const {
+        reason, message, type, involvedObject: { name, namespace },
+      } = req.body;
+      logger.log(`[${reason}] received ${type} kubernetes event: ${message} ${name} in ${namespace}`);
+
+      // remove only pods in your namespace and due to backoff errors
+      if ((namespace.match('^pipeline-.*') || namespace.match('^worker-.*'))
```
aerlaut, line 21: This string includes regex characters `^` and `.*`. Instead of a string, would it be clearer to change this string as a Regex? e.g. `/^pipeline-.*/`

src/api.v2/routes/kubernetes.js
```diff
@@ -0,0 +1,33 @@
...
+       && reason === 'BackOff' && type !== 'Normal' && message.includes('restarting')) {
+        const k8sApi = kc.makeApiClient(k8s.CoreV1Api);
+        logger.log(`removing pod ${name} in ${namespace}`);
+        await k8sApi.deleteNamespacedPod(name, namespace);
+      }
+      res.status(200).send('ok');
```
aerlaut: We have the `OK()` helper in the API (`src/utis/responses/OK`). Can you use that instead of hard-coding this?

src/api.v2/routes/kubernetes.js
kafkasl: what's the difference? I think they do the same

src/api.v2/routes/kubernetes.js
kafkasl: I prefer not to, they are not equivalent and the `OK()` should still be included so it would be `res.json(OK())` because we are in `routes` instead of `route-services` where the naked `OK()` is used. I had a lot of issues with the `event-exporter` due to the responses we were sending to it, so I'd rather leave it as it is considering it's an isolated and exceptional endpoint.

kafkasl: I modified it in the end, we'll need to verify it works correctly with the `event-exporter` because unfortunately it can't be tested until it's merged.

src/api.v2/routes/kubernetes.js
aerlaut: I think it just makes it clearer that you're matching against a regex, not a string. It wouldn't affect the functionality, but it's fine if it's kept this way.

aerlaut (approved):

PR #359, opened 2022-05-27
[BIOMAGE-1877] Implement run qc for api v2

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1877

https://ui-martinfosco-ui715-api3-v2.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/iac/pull/405
https://github.com/hms-dbmi-cellenics/ui/pull/715
https://github.com/hms-dbmi-cellenics/pipeline/pull/237

Dynamodb migration PR (check [this thread](https://github.com/hms-dbmi-cellenics/api/pull/359#discussion_r883704328))
https://github.com/hms-dbmi-cellenics/iac/pull/406

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/sql/knexfile.js
```diff
@@ -3,6 +3,11 @@ const _ = require('lodash');
 const config = require('../config');
 const getConnectionParams = require('./getConnectionParams');
 
+// This was necessary because all of processingConfig is camelcase except for
+// absolute_theshold, which breaks the camelcase we had assumed
+// Consider migrating the property absolute_threshold to absoluteThreshold instead of doing this
+const processingConfigExceptions = ['methodSettings'];
```
cosa65: The solution proposed in the comment is probably better than what I did which is a fast patch, should I do that one instead? Some opinions on this pls. 
If we do it, it will probably not be too complicated.

In fact we could also add a patch make sure that anything stored as json is fully camelcase to avoid inconsistencies like this one creeping in

ivababukova: This patch works, but it is definitely less clean than having absolute threshold in camel case. My opinion is that the best solution is to have absolute threshold in camel case and thus not have any exceptions. Exceptions can only be a cause of problems and create technical debt long term.

I would use this patch you coded in here if there are other fields like "absolute threshold" and/or if it would be too time consuming and complicated to make the exceptions stop being exceptions and use camel case. Or if you just want to merge this PR and have this patch temporary and make absolute threshold to camel case in a separate PR.

ivababukova: Overall, this patch is not too bad, it will work and if you decide to go with it, it is fine by me. Still, my opinion is that whenever we can achieve a cleaner solution with a reasonable amount of effort, we should strive for the cleaner solution.

src/api.v2/routes/qc.js
```diff
@@ -1,4 +1,17 @@
...
+  'qc#get': [
+    expressAuthorizationMiddleware,
+    (req, res, next) => getQCState(req, res).catch(next),
+  ],
+  'qc#run': [
+    expressAuthorizationMiddleware,
+    (req, res, next) => runQC(req, res).catch(next),
+  ],
+  'qc#response': (req, res, next) => {
+    handleResponse(req, res).catch(next);
```
ivababukova, line 11: why is it ok to not call the authorization middleware in this case?

cosa65: they are sns messages, we don't create them.

This was already like this in v1, and I think that is because `parseSNSMessage` calls `validator.validate` from this lbirary https://github.com/aws/aws-js-sns-message-validator

cosa65: we could add a ticket to check that this is actually secure (not in data model epic though, I think it is separate and already exists)

src/sql/knexfile.js
cosa65: ok, I'll try to do that then, I'll add a PR for the dynamoToSql script in iac and modify the pipeline, the ui and api, will leave links here when it is ready

src/api.v2/routes/qc.js
ivababukova: yes, I think we should evaluate that.

ivababukova: and yes, it is outside of scope for data model

src/sql/knexfile.js
cosa65: Unrelated to this PR, but I added [this](https://github.com/hms-dbmi-cellenics/iac/pull/405) to clean up the processing configs in production, it seems they still have the `api_url` and `auth_JWT` properties which were going to be removed in a previous ticket

cosa65: Also I added [this script](https://github.com/hms-dbmi-cellenics/iac/pull/406/files) to run the migration in dynamo to change to camelcase the processing config, I tested that v2 works correctly with a migrated experiment locally, but I will perform more testing on this script when it is time to run it (so when merging this PR and when running it in production for the release too)

src/api.v2/controllers/qcController.js
```diff
@@ -0,0 +1,64 @@
...
+
+const { createQCPipeline } = require('../helpers/pipeline/pipelineConstruct');
+const handleQCResponse = require('../helpers/pipeline/handleQCResponse');
+
+const getLogger = require('../../utils/getLogger');
+const parseSNSMessage = require('../../utils/parse-sns-message');
+
+const logger = getLogger('[QCController] - ');
+
+const runQC = async (req, res) => {
```
ivababukova, line 14: I was looking and comparing this `qcController` with `gem2sController`. I can see that in some respect, they are almost the same. For example, the handleResponse function in each of them is almost identical. That makes a lot of sense, what I don't understand is why there are differences between what `runQC` and what `runGem2s` return? Why for `runGem2s` it's fine to just have `res.json(OK())` and for `runQC` we have to return the `stateMachineArn` and `executionArn`? Shouldn't we be returning those also for runGem2s?

src/api.v2/controllers/qcController.js
```diff
@@ -0,0 +1,64 @@
...
+
+  const { io, parsedMessage } = result;
+
+  const isSnsNotification = parsedMessage !== undefined;
+  if (isSnsNotification) {
+    try {
+      await handleQCResponse(io, parsedMessage);
+    } catch (e) {
+      logger.error(
+        'pipeline response handler failed with error: ', e,
```
ivababukova: for consistency with the log in gem2s handleResponse function, you can add `qc` before `pipeline` in the log message.

src/api.v2/helpers/pipeline/gem2s.js
```diff
@@ -19,17 +19,35 @@ const logger = getLogger('[Gem2sService] - ');
...
+
+  await new ExperimentExecution().upsert(
+    {
+      experiment_id: experimentId,
+      pipeline_type: 'qc',
+    },
+    {
+      state_machine_arn: stateMachineArn,
+      execution_arn: executionArn,
+    },
```
ivababukova: why do we not need to call `ExperimentExecution` for gem2s?

src/api.v2/helpers/pipeline/handleQCResponse.js
```diff
@@ -0,0 +1,147 @@
...
+  if (output.plotDataKeys) {
+    const plotConfigUploads = Object.entries(output.plotDataKeys)
+      .map(async ([plotUuid, s3DataKey]) => (
+        await new Plot().upsert(
+          { id: plotUuid, experiment_id: experimentId },
+          { s3_data_key: s3DataKey },
+        )
+      ));
+
+    logger.log('Uploading plotData for', taskName, 'to DynamoDB');
```
ivababukova: dynamo?

src/api.v2/helpers/pipeline/handleQCResponse.js
```diff
@@ -0,0 +1,147 @@
...
+      plotConfigUploads.map((p) => p.catch((e) => logger.error(e))),
+    );
+  }
+};
+
+const updateProcessingConfig = async (taskName, experimentId, output, sampleUuid) => {
+  // TODO the processing config validation was not being enforced because the 'anyOf' requirement
+  // was not being correctly applied. This needs to be refactored together with the
+  // pipeline and ideally while unifying the qc & gem2s responses.
+  // await validateRequest(output, 'ProcessingConfigBodies.v1.yaml');
```
ivababukova, line 63: do we have a ticket for this work?

src/api.v2/helpers/pipeline/gem2s.js
```diff
@@ -19,17 +19,35 @@ const logger = getLogger('[Gem2sService] - ');
...
+  );
 
-const continueToQC = () => {
+  logger.log('Started qc successfully');
 };
 
 hookRunner.register('uploadToAWS', [
```
ivababukova, line 39: I know that this is not directly part of your PR, but what is the point of these hooks? Why calling `hookRunner.register` and `hookRunner.registerAll` in the middle of this file? When will `hookRunner.register` and `hookRunner.registerAll` run? Every time we import the file? Or when? 
I am finding them confusing, also I notice that they are used only in this file and in `handleQCresponse`, but they are still super generic. Can you explain me what is the point of having this over-generalisation?

src/api.v2/helpers/pipeline/gem2s.js
```diff
@@ -19,17 +19,35 @@ const logger = getLogger('[Gem2sService] - ');
 
 const hookRunner = new HookRunner();
 
-const saveProcessingConfigFromGem2s = ({ experimentId, item }) => {
-  logger.log('Saving processing config for gem2s');
-  new Experiment().updateById(experimentId, { processing_config: item });
-  logger.log('Finished saving processing config for gem2s');
-};
+const continueToQC = async (payload) => {
```
ivababukova, line 22: shouldn't `continueToQC` actually be called `startQC`?

src/api.v2/helpers/pipeline/gem2s.js
```diff
@@ -19,17 +19,35 @@ const logger = getLogger('[Gem2sService] - ');
...
+    },
+  );
 
-const continueToQC = () => {
+  logger.log('Started qc successfully');
 };
 
 hookRunner.register('uploadToAWS', [
-  saveProcessingConfigFromGem2s,
   continueToQC,
```
ivababukova, line 40: It would be nice if you renamed `gem2sResponse` to `handleGem2sResponse`, to make the naming consistent with gem2s. Same with `gem2sCreate` and `createQCPipeline`. Having consitent naming can really help with understanding the code easier

src/api.v2/helpers/pipeline/handleQCResponse.js
```diff
@@ -0,0 +1,147 @@
...
+  AWSXRay.getSegment().addMetadata('message', message);
+
+  await validateRequest(message, 'PipelineResponse.v1.yaml');
+
+  await hookRunner.run(message);
+
+  const { experimentId } = message;
+  const { error = false } = message.response || {};
+
+  let output = null;
```
ivababukova: could you rename this `output` to `s3output` or `s3data`? Something less generic to indicate from its name immediately what it is about.

src/api.v2/helpers/pipeline/handleQCResponse.js
```diff
@@ -0,0 +1,147 @@
...
+  // pipeline and ideally while unifying the qc & gem2s responses.
+  // await validateRequest(output, 'ProcessingConfigBodies.v1.yaml');
+
+  const experiment = new Experiment();
+
+  const {
+    processingConfig: previousConfig,
+  } = await experiment.findById(experimentId).first();
+
+  if (sampleUuid !== '') {
```
ivababukova, line 71: can you give me a valid scenario when `sampleUuid === ''` and we want to be doing what is in the `else` block? When and why `sampleUuid === ''`?

src/api.v2/helpers/pipeline/handleQCResponse.js
```diff
@@ -0,0 +1,147 @@
...
+  const { error = false } = message.response || {};
+
+  let output = null;
+  // if there aren't errors proceed with the updates
+  if (!error && 'output' in message) {
+    const { input: { sampleUuid, taskName } } = message;
+
+    output = await getS3Output(message);
+
+    await updatePlotData(taskName, experimentId, output);
```
ivababukova: What about:
instead of checking inside `updatePlotData` if `output.plotDataKeys` is there or not, and then passing the whole `output` as an argument to `updatePlotData`, 
you do the following:
check if `output.plotDataKeys` exists before calling `updatePlotData`. If it exists, then call `updatePlotData`, pass only `plotDataKeys` as a parameter, not the entire `output` and remove the umbrella if stmnt in `updatePlotData`?

src/api.v2/helpers/pipeline/handleQCResponse.js
```diff
@@ -0,0 +1,147 @@
...
+const updateProcessingConfig = async (taskName, experimentId, output, sampleUuid) => {
+  // TODO the processing config validation was not being enforced because the 'anyOf' requirement
+  // was not being correctly applied. This needs to be refactored together with the
+  // pipeline and ideally while unifying the qc & gem2s responses.
+  // await validateRequest(output, 'ProcessingConfigBodies.v1.yaml');
+
+  const experiment = new Experiment();
+
+  const {
+    processingConfig: previousConfig,
```
ivababukova, line 68: previous config? Why are we updating to previous config?

src/api.v2/helpers/pipeline/handleQCResponse.js
```diff
@@ -0,0 +1,147 @@
...
+  const statusResToSend = { pipeline: statusRes[constants.QC_PROCESS_NAME] };
+
+  // Concatenate into a proper response.
+  const response = {
+    ...message,
+    status: statusResToSend,
+    type: constants.QC_PROCESS_NAME,
+  };
+
+  if (output !== null) {
```
ivababukova, line 112: can `output` be null? Based on my code reading, if `output` is null, then `if (output.plotDataKeys)` inside function `updatePlotData` will error out. Or am I missing something?

src/api.v2/controllers/qcController.js
cosa65: You're right, I checked and runPipeline doesn't use the arns even for v1, so I'll make qc also return `OK()`

src/api.v2/helpers/pipeline/gem2s.js
cosa65: because it's in the hook `const continueToQC = async`, which triggers a new qc run.

Oh, I had refactored this, the upsert is already being done inside `createQCPipeline` so I can remove this

src/api.v2/helpers/pipeline/handleQCResponse.js
cosa65: I am not sure, this comment existed from before my changes

src/api.v2/helpers/pipeline/gem2s.js
cosa65: they are run when the api is created only I believe

cosa65: I don't know about the reason for the hooks system compared to other options because I didn't take part in their creation.

The benefit I can see is that there were some things that had to be done when some of the step functions (qc/gem2s) steps finished, this is a good way to specify for which steps finishing something should be run (compared to using ifs this is more readable).

src/api.v2/helpers/pipeline/gem2s.js
cosa65: I like `continueToQC` becuase it makes it clear that we have kind of concatenated qc after gem2s

src/api.v2/helpers/pipeline/gem2s.js
cosa65: Sounds good, I did a lot of copy pasting without a lot of refactoring so I didn't look into fixing these issues

src/api.v2/helpers/pipeline/gem2s.js
ivababukova: I see. To make this more readable, can you move all the hooks register and registerAll calls to happen at the top of the file, similarly to how it is done for `handleQCresponse`? It will be much clearer that way

src/api.v2/helpers/pipeline/handleQCResponse.js
cosa65: did some renaming in the functions involved

src/api.v2/helpers/pipeline/handleQCResponse.js
cosa65: I am not sure honestly, I think it is legacy code from when we used the processing config we store without an assigned sample id

src/api.v2/helpers/pipeline/handleQCResponse.js
cosa65: because we receive a config update, we merge both and upload the new config.

src/api.v2/helpers/pipeline/handleQCResponse.js
cosa65: I could make it `output && output.plotDataKeys`

src/api.v2/helpers/pipeline/handleQCResponse.js
cosa65: I feel like it would make `handleQCResponse` less readable because it would have more of a mix of code from different concerns.

As an alternative, I could rename `updatePlotData` to `updatePlotDataIfNecessary` maybe

ivababukova: I was suggesting this, because you then pass less things to `updatePlotData` as parameters. The name of the function is ok as is in my opinion.

src/api.v2/helpers/pipeline/handleQCResponse.js
ivababukova: let's make a ticket for removing this and the other legacy parts.

src/api.v2/helpers/pipeline/gem2s.js
ivababukova: that's fair. I was reading in all the code, and that's why I noticed this

ivababukova (approved):
lgtm after testing in staging with v1 and v2

PR #377, opened 2022-06-21
1635 Add example experiments creation

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1635

https://ui-martinfosco-ui747-api377.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/747

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api.v2/controllers/experimentController.js
```diff
@@ -129,11 +137,41 @@ const downloadData = async (req, res) => {
...
+const cloneExperiment = async (req, res) => {
+  const {
+    params: { experimentId: fromExperimentId, toExperimentId },
+  } = req;
+
+  logger.log(`Cloning experiment ${fromExperimentId} into ${toExperimentId}`);
+
+  // toExperiment might have metadata tracks,
+  // we want to start from an empty one so clear them out
+  await new MetadataTrack().delete({ experiment_id: toExperimentId });
```
aerlaut: `toExperimentId` shouldn't exist in the database right? So why do we need to delete this?

src/api.v2/model/Sample.js
```diff
@@ -38,6 +42,11 @@ class Sample extends BasicModel {
     const sampleFileFields = ['sample_file_type', 'size', 'upload_status', 's3_path'];
     const sampleFileFieldsWithAlias = sampleFileFields.map((field) => `sf.${field}`);
     const fileObjectFormatted = sampleFileFields.map((field) => [`'${field}'`, field]);
+
+    // Add sample file id (needs to go separate to avoid conflict with sample id)
+    sampleFileFieldsWithAlias.push('sf.id as sf_id');
+    fileObjectFormatted.push(['\'id\'', 'sf_id']);
```
aerlaut, line 48: Where is this used? is this a fix for the query below?

src/api.v2/controllers/experimentController.js
cosa65: it exists because in the ticket it says we should clone an example experiment into the empty experiment that causes us to show the example experiments suggestion
image

src/api.v2/model/Sample.js
cosa65: `getSamples` was returning every column I needed from the sample files except for the sample file ids, so I thought I could just add the sample file ids to be returned here too

src/api.v2/controllers/experimentController.js
```diff
@@ -153,11 +154,16 @@ const cloneExperiment = async (req, res) => {
   // we want to start without any old tracks so clear them out
   await new MetadataTrack().delete({ experiment_id: toExperimentId });
 
-  // Get samples order so we preserve the relative order of the original samples in the cloned ones
-  const { samplesOrder } = await new Experiment().findById(fromExperimentId).first();
+  let samplesToCloneIds = samplesSubsetIds;
```
aerlaut: Can you rename this during import at line 148 above? (i.e. `body: { samplesSubsetIds : samplesToCloneIds = null }`), example: https://stackoverflow.com/a/56700589/1940886

src/api.v2/controllers/experimentController.js
```diff
@@ -8,33 +8,43 @@ const { OK, NotFoundError } = require('../../utils/responses');
...
 
 const getExperimentBackendStatus = require('../helpers/backendStatus/getExperimentBackendStatus');
+const Sample = require('../model/Sample');
+const MetadataTrack = require('../model/MetadataTrack');
 
 const logger = getLogger('[ExperimentController] - ');
 
 const getAllExperiments = async (req, res) => {
   const { user: { sub: userId } } = req;
+  console.log(`Getting all experiments for user: ${userId}`);
```
kafkasl: remove the console logs, or move them into `logger.log` if you wanna keep them

src/api.v2/controllers/experimentController.js
```diff
@@ -129,11 +138,45 @@ const downloadData = async (req, res) => {
...
+
+const cloneExperiment = async (req, res) => {
+  const getAllSampleIds = async (experimentId) => {
+    const { samplesOrder } = await new Experiment().findById(experimentId).first();
+    return samplesOrder;
+  };
+
+  const {
+    params: { experimentId: fromExperimentId, toExperimentId },
+    body: { samplesSubsetIds = await getAllSampleIds(fromExperimentId) },
```
kafkasl: `samplesSubsetIds` is not very clear to me, maybe `samplesToClone` ?

src/api.v2/controllers/experimentController.js
```diff
@@ -129,11 +138,45 @@ const downloadData = async (req, res) => {
...
 
+
+const cloneExperiment = async (req, res) => {
+  const getAllSampleIds = async (experimentId) => {
+    const { samplesOrder } = await new Experiment().findById(experimentId).first();
+    return samplesOrder;
+  };
+
+  const {
+    params: { experimentId: fromExperimentId, toExperimentId },
```
kafkasl: is the new experiment then created in the UI? because I assume it would be created by the API, but maybe it makes more sense like this. How is it going to work?

src/api.v2/controllers/experimentController.js
```diff
@@ -129,11 +138,45 @@ const downloadData = async (req, res) => {
...
+    body: { samplesSubsetIds = await getAllSampleIds(fromExperimentId) },
+  } = req;
+
+  logger.log(`Cloning experiment ${fromExperimentId} into ${toExperimentId}`);
+
+  // toExperiment might be an old experiment with metadata tracks,
```
kafkasl: I think this is kind of risky, like reusing old experiment also not sure how you could "select" an old experiment in the UI. In my opinion we should always create a new copy from scratch when you are cloning an experiment. I don't see any benefit in reusing old experiments, but potentially I see many issues (if we don't clean it up correctly)

src/api.v2/model/Experiment.js
```diff
@@ -43,20 +44,26 @@ class Experiment extends BasicModel {
     ];
 
     const aliasedExperimentFields = fields.map((field) => `e.${field}`);
-    function mainQuery() {
-      this.select([...aliasedExperimentFields, 'm.key'])
+
+    const result = await collapseKeyIntoArray(
```
kafkasl, line 57: not sure if this is an improvement, it makes it harder to read the actual query, and harder to figure out which are the parameters to `collapseKeyIntoArray`

src/api.v2/model/Sample.js
```diff
@@ -81,6 +90,71 @@ class Sample extends BasicModel {
       );
     });
   }
+
+  /**
+   * Creates copy samples from one experiment to another one
```
kafkasl: creates copy samples? not sure what it means:
* creates a copy
* copies samples

src/api.v2/controllers/experimentController.js
```diff
@@ -8,33 +8,43 @@ const { OK, NotFoundError } = require('../../utils/responses');
...
   const { user: { sub: userId } } = req;
+  console.log(`Getting all experiments for user: ${userId}`);
 
   const data = await new Experiment().getAllExperiments(userId);
 
+  console.log(`Finished getting all experiments for user: ${userId}, length: ${data.length}`);
+  res.json(data);
+};
+
+const getAllExampleExperiments = async (req, res) => {
```
kafkasl: in the end we agreed we won't have distinction between public & example right? because then this function should probably be called `getPublicExperiments`. (btw I think the `All` is rarely needed, you just want the experiments matching something, makes the name unnecessarily longer IMO)

src/specs/api.v2.yaml
```diff
@@ -1154,7 +1186,56 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/HTTPError'
-
+  '/experiments/{experimentId}/clone/{toExperimentId}':
```
kafkasl: I'm not clear where this `toExperimentId` would come. I think it should be the api creating it

src/api.v2/controllers/experimentController.js
cosa65: Based on what the ticket said I understood we created an empty experiment, the experiment showed us the optional datasets and then the samples get cloned into this empty experiment. So the experiment is already created because that's where we get the options of example datasets
image

src/api.v2/controllers/experimentController.js
cosa65: Same as https://github.com/hms-dbmi-cellenics/api/pull/377#discussion_r904927722

src/api.v2/model/Experiment.js
cosa65: The main change is to stop using mainQuery as a function, I will keep that change but I can make `const mainQuery = ...`

src/specs/api.v2.yaml
cosa65: Same as https://github.com/hms-dbmi-cellenics/api/pull/377#discussion_r904927722

src/api.v2/model/Sample.js
```diff
@@ -81,6 +90,71 @@ class Sample extends BasicModel {
...
+   *
+   * @param {*} fromExperimentId
+   * @param {*} toExperimentId
+   */
+  async copyTo(fromExperimentId, toExperimentId, samplesOrder) {
+    const result = await this.getSamples(fromExperimentId);
```
StefanBabukov: Maybe call this var samples or fromSamples?

src/api.v2/model/Sample.js
```diff
@@ -81,6 +90,71 @@ class Sample extends BasicModel {
...
+        const sample = result.find(({ id }) => id === fromSampleId);
+
+        const toSampleId = uuidv4();
+
+        newSampleIds.push(toSampleId);
+
+        sampleRows.push({
+          id: toSampleId,
+          experiment_id: toExperimentId,
+          name: sample.name,
```
StefanBabukov, line 131: you can just use `...sample` here

src/api.v2/model/Sample.js
cosa65: Renamed to `fromSamples`

src/api.v2/controllers/experimentController.js
cosa65: Discussed and alreayd adapted the code to what we resolved, create a new experiment and leave this experiment empty

src/api.v2/controllers/experimentController.js
```diff
@@ -129,11 +137,51 @@ const downloadData = async (req, res) => {
...
+
+  const clonedSamplesOrder = await new Sample()
+    .copyTo(fromExperimentId, toExperimentId, samplesToCloneIds);
+
+  await new Experiment().updateById(
+    toExperimentId,
+    { samples_order: JSON.stringify(clonedSamplesOrder) },
+  );
+
+  logger.log(`Finished cloning experiment ${fromExperimentId}, new expeirment's id is ${toExperimentId}`);
```
kafkasl: typo in the log comment

src/api.v2/controllers/experimentController.js
cosa65: Right now this endpoint gets example experiments, and it is being used in this way in the UI too, if/when we start using public experiments is when I think renaming this would be better I think

cosa65: Renamed to `getExampleExperiments` btw

src/api.v2/controllers/experimentController.js
cosa65: Renamed to `samplesToCloneIds`

kafkasl (approved):

PR #442, opened 2023-07-19
Merge

cosa65:
# Description

# Details
N/A

https://github.com/hms-dbmi-cellenics/pipeline/pull/312
https://github.com/hms-dbmi-cellenics/worker/pull/307
https://github.com/hms-dbmi-cellenics/api/pull/442
https://github.com/hms-dbmi-cellenics/ui/pull/848

#### Integration test branch
master

# Merge checklist

### Code updates
- [x] All new dependency licenses have been checked for compatibility

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/config/default-config.js
```diff
@@ -1,10 +1,10 @@
 const AWS = require('aws-sdk');
 const getLogger = require('../utils/getLogger');
+const { ACCOUNT_ID } = require('../api.v2/constants');
 
 const logger = getLogger();
-const { ADMIN_SUB } = require('../api.v2/constants');
 
-const githubOrganisationName = 'hms-dbmi-cellenics';
+const githubOrganisationName = process.env.AWS_ACCOUNT_ID === ACCOUNT_ID.HMS ? 'hms-dbmi-cellenics' : 'biomage-org';
```
alexvpickering, line 7:
```suggestion
const githubOrgByAccountId = {
  [ACCOUNT_ID.HMS]: 'hms-dbmi-cellenics',
  [ACCOUNT_ID.BIOMAGE]: 'biomage-org',
};
const githubOrganisationName = githubOrgByAccountId[process.env.AWS_ACCOUNT_ID];
```

src/utils/getAdminSub.js
```diff
@@ -1,7 +1,8 @@
 const config = require('../config');
 const getAwsPoolId = require('../api.v2/helpers/cognito/getAwsPoolId');
+const { ACCOUNT_ID } = require('../api.v2/constants');
 
-const adminEmail = 'cellenics_admin@listserv.med.harvard.edu';
+const adminEmail = process.env.AWS_ACCOUNT_ID === ACCOUNT_ID.HMS ? 'cellenics_admin@listserv.med.harvard.edu' : 'admin@biomage.net';
```
alexvpickering: would it make sense to have this kind of logic in a single location (like `domainSpecificContent.js` for the ui)?

src/config/default-config.js
alexvpickering: another potential candidate for moving to a single domain specific content

src/api.v2/middlewares/authMiddlewares.js
```diff
@@ -30,7 +32,17 @@ const checkUserAuthenticated = (req, next) => {
...
+const checkForPrivacyPolicyAgreement = (req, next) => {
+  const isBiomageDeployment = BIOMAGE_DOMAIN_NAMES.includes(config.domainName);
 
+  if (req.user['custom:agreed_terms'] !== 'true' && isBiomageDeployment) {
+    next(new NotAgreedToTermsError('The user hasnt agreed to the privacy policy yet.'));
+    return false;
+  }
+
+  return true;
+};
```
alexvpickering, line 45: this could be a candidate for another domain specific content as it is something that is only used for the Biomage deployment

src/api.v2/middlewares/authMiddlewares.js
alexvpickering, line 1: additions to this file could be a candidate for another domain specific content as it is something that is only used for the Biomage deployment

src/utils/constants.js
```diff
@@ -25,6 +25,7 @@ const EXECUTION_DOES_NOT_EXIST = 'ExecutionDoesNotExist';
 const ACCESS_DENIED = 'AccessDeniedException';
 
 const PUBLIC_ACCESS_ID = '00000000-0000-0000-0000-000000000000';
+const BIOMAGE_DOMAIN_NAMES = ['scp.biomage.net', 'scp-staging.biomage.net'];
```
alexvpickering: these are currently only used for checking if Biomage's privacy policy has been accepted. I would not really consider these as general purpose constants and should probably be in a domain specific file

src/config/test-config.js
```diff
@@ -1,21 +1,19 @@
...
   port: 3000,
   clusterEnv: 'test',
   awsAccountId: '000000000000',
   awsRegion: 'eu-west-1',
-  corsOriginUrl: ['https://test.url2.com', 'https://test.url1.com', 'https://scp.biomage.net'],
-  emailDomainName: 'https://scp.biomage.net',
-  domainName: 'scp.biomage.net',
+  corsOriginUrl: ['https://test.url2.com', 'https://test.url1.com', 'https://test.com'],
+  emailDomainName: 'https://test.com',
+  domainName: 'test.com',
```
alexvpickering, line 10: I suggest that we make the test configuration domain agnostic -- all the tests still pass with updated snaps

tests/config/__snapshots__/default-config-staging.test.js.snap
```diff
@@ -25,7 +24,7 @@ Object {
   "domainName": "scp-staging.biomage.net",
   "emailDomainName": "https://ui-default.scp-staging.biomage.net",
   "pipelineIgnoreSSLCertificate": false,
-  "pipelineInstanceConfigUrl": "https://raw.githubusercontent.com/hms-dbmi-cellenics/releases/master/staging/pipeline.yaml",
+  "pipelineInstanceConfigUrl": "https://raw.githubusercontent.com/biomage-org/releases/master/staging/pipeline.yaml",
```
alexvpickering: All these snaps contain references to `biomage-org` because the default `githubOrganisationName` is  `biomage-org` (as well as a few hardcoded values in `tests/config/default-config-staging.test.js`). Is this a requirement so that the deployment works with all Biomage's private deployments or can some of this organisation name spread be eliminated?

tests/utils/__snapshots__/getAdminSub.test.js.snap
```diff
@@ -7,7 +7,7 @@ Array [
     "UserAttributes": Array [
       Object {
         "Name": "email",
-        "Value": "cellenics_admin@listserv.med.harvard.edu",
+        "Value": "admin@biomage.net",
```
alexvpickering: similar question as above - is it reasonable to eliminate the need for organisation name within these snaps

alexvpickering (changes requested):
looks good - most comments are aimed at keeping organisation names to a minimum and the desire to centralize organisation specific code and configuration so that it is easy to find for when someone wants to join in our efforts

tests/config/default-config-staging.test.js
```diff
@@ -28,11 +28,11 @@ describe('default-config', () => {
     process.env.AWS_DEFAULT_REGION = 'eu-west-1';
     process.env.AWS_REGION = 'eu-west-1';
     process.env.RDS_SANDBOX_ID = 'default';
-    process.env.AWS_ROLE_ARN = 'arn:aws:iam::242905224710:role/api-role-production';
+    process.env.AWS_ROLE_ARN = 'arn:aws:iam::123456789:role/api-role-production';
```
alexvpickering: I changed these hardcoded values for the tests as there doesn't seem to be any need for them to be organisation related

tests/config/__snapshots__/default-config-staging.test.js.snap
cosa65: Maybe moving everything domain specific into `getDomainSpecific()` we can just mock this function and make the returns be domain agnostic

src/config/default-config.js
cosa65: This one is where all the variables are extracted from (including the account id, etc) so it is a little bit harder to try to move it into `getDomainSpecific`  (because `getDomainSpecific` checks this file to know which domain it is in, so we would have to figure out a way to avoid a circular dependency). 

I think ideally we could try to keep the domain specific checks down to this file and `getDomainSpecific` (similarly to how in the ui we. have `next.config.js` and `getDomainSpecificContent`)

src/api.v2/middlewares/authMiddlewares.js
cosa65: I moved `checkForPrivacyPolicyAgreement` to `getDomainSpecific`, let me know if you meant anytihng else in this file too

tests/config/default-config-staging.test.js
cosa65: if it's only test hcanges I think it sohuld be ok, but if there is any changes we want to perform that affect the platform itself it might be a good idea to raise the PR first in biomage-org and htem merge from there to this branch so that we don't end up with very divergent repos.

The idea (as I understood it) of these PRs is that we are merging what is in our biomage-org `master` to hms's `master`, direct changes to this branch make it no longer part of biomage-org `master`)

src/config/getDomainSpecificContent.js
alexvpickering, line 1: This is a great - thank you! There are a few other items that I suggest should be moved into here:

- `buildPipelineStatusEmailBody.js` is checking if using HMS account to determine notification email sender
- `buildUserInvitedEmailBody.js` and `buildUserInvitedNotRegisteredEmailBody` are both adding additional content to the email for non-HMS deployments

There are also a few test files that run tests separately for both `ACCOUNT_ID.BIOMAGE` and `ACCOUNT_ID.HMS`  (`default-config-production.test.js`) or only `ACCOUNT_ID.HMS` (`buildPipelineStatusEmailBody.test.js`, `buildUserInvitedEmailBody.test.js`, and `buildUserInvitedNotRegisteredEmailBody.test.js`) and then check against snapshots. I think that these tests (more so than the `ui` ones we previously discussed) are good candidates to just loop through the keys (AWS account ids) of `domainSpecificContent`, run the test, and check against the snapshot. Doing this will remove duplicate code from these test files as well as avoid inserting organization specific tests everywhere that become silly to add to if additional parties want to join.

tests/config/default-config-production.test.js
```diff
@@ -22,10 +23,11 @@ describe('default-config', () => {
 
   it('Returns correct values for BIOMAGE production', () => {
     const prodEnvironment = 'production';
```
alexvpickering: see comment above -- I think this could be one test ala `it('Return correct values for production environments')` and then loop through all keys in `domainSpecificContent`.

tests/config/default-config-staging.test.js
alexvpickering: apologies for that -- I won't commit to these set of PRs directly

src/config/getDomainSpecificContent.js
```diff
@@ -0,0 +1,33 @@
+const { ACCOUNT_ID } = require('../api.v2/constants');
+
+const domainSpecificContent = {
+  test: {
+    githubOrganisationName: 'org',
+    adminEmail: 'admin@example.com',
+    enforcePrivacyPolicyAgreement: true,
```
alexvpickering: As it is, local development is broken for me (I log in to HMS staging with no privacy policy)

```suggestion
    enforcePrivacyPolicyAgreement: false,
```

StefanBabukov: This is for the test environment, it shouldnt change anything for your local development, did you run the api and UI with `DEV_ACCOUNT=HMS npm start` ?

src/config/getDomainSpecificContent.js
StefanBabukov: Your approach makes sense but in the future should not be done like this. The tests should be isolated to the function they are testing as much as possible. So ideally we have 1 test for `getDomainSpecificContent` where we check if all the values are returned correctly depending on the account and then the other tests such as `buildUserInvitedNotRegisteredEmailBody` don't check the values and loop through accounts from `getDomainSpecific`, because those values are already tested in the `getDomainSpecificContent` test. Instead they check if they are getting those values (doesn't matter what the value is) from `getDomainSpecificContent`. Keeping the isolation of concerns will make easier to develop and test in the future. I'm not sure if that makes sense.

 I still added the loops for now, since it wasn't hard to do and i realised this when i already pushed.

alexvpickering: Makes a lot of sense -- I like that approach a lot and it accomplishes a similar outcome of not having multiple organization specific tests. Definitely feel free to change the tests to reflect the isolation of concerns.

src/config/getDomainSpecificContent.js
alexvpickering: `DEV_ACCOUNT` is not used by the `api` AFAIK.

Look down at `getDomainSpecificContent`: it checks for `AWS_ACCOUNT_ID` which for development needs to be unset or `'000000000000'` (the default that is used if unset -- see `default-config.js`). As a result, `getDomainSpecificContent` returns `domainSpecificContent.test` and I can't develop locally due to complaints from the API about not agreeing to privacy policy.

`DEV_ACCOUNT=HMS npm start` is run in the ui. I tried also doing in the api but nothing different (also as mentioned `DEV_ACCOUNT` isn't anywhere in the api code base). 

I also tried setting `AWS_ACCOUNT_ID` to HMS appropriate value, which fixes the privacy policy issue but creates local infrastructure communication issues (cant upload data) because the localhost aws account needs to be `'000000000000'`.

alexvpickering (approved):

tests/config/getDomainSpecificContent.test.js
```diff
@@ -0,0 +1,35 @@
...
+  it('check content for the biomage account', () => {
+    process.env.AWS_ACCOUNT_ID = 'some_private_account_id';
+    process.env.NODE_ENV = 'production'; // Since the function checks if env is not 'test'
+
+    const defaultContent = getDomainSpecificContent();
+    expect(defaultContent).toMatchSnapshot('getDomainSpecificContent for biomage private account');
+  });
+
+  // Test for the 'test' account
+  it('check content for the biomage account', () => {
```
alexvpickering:
```suggestion
  it('check content for the test account', () => {
```

PR #484, opened 2023-09-26
Push migrations only in hms-dbmi-cellenics repo

kafkasl:
# Description

Currently, we are pushing to `iac` from the api. When this is done in a fork, it means that the `iac` repo receives the same updates twice: once from syncing the `iac` repo, and the second when releasing the forked `api` repo. In order to prevent this situation, I propose to push only from the original `api` repository. However, this leads to a second problem due to this behavior:
1. When PRs are merged to the api, migrations are pushed in to `iac/sql/staging` migrations
2. When the `api` is released, migrations are pushed to `iac/sql/production`. 

This means that fork repos won't be able to sync production migrations (from `hms-dbmi-cellenics/iac` to `fork/iac`) unless the original repo has been released to production (point 2). 

To solve this I propose to just get rid of the two different migrations folders `staging`, `production` [done in this `iac` PR](https://github.com/hms-dbmi-cellenics/iac/pull/562). We already have a migration script in `iac` to choose whether to migrate staging or production conveniently. 

This solution would require 0 modifications of our current development flow because the migrations in the api remain in the same place, and the github action in `iac` takes care of the different source location.

The only downside will be not seeing in the folders which migrations have been applied. I believe this is not an issue as long as everyone takes ownership of releasing and migrating their own features. It is also possible to check which migrations have been applied using the command line, tablePlus or any other method to inspect the DB. If it's inconvenient we could create a very short script to run `knex migrate:list` in `iac` repo. 

# Details
#### URL to issue
https://github.com/hms-dbmi-cellenics/issues/issues/53

N/A

https://github.com/hms-dbmi-cellenics/iac/pull/562
https://github.com/hms-dbmi-cellenics/api/pull/484

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


ivababukova:
Hi @kafkasl, sounds like a neat and easy to implement proposal. I can't think of anything better to do at this point.

About this comment:

```The only downside will be not seeing in the folders which migrations have been applied. I believe this is not an issue as long as everyone takes ownership of releasing and migrating their own features.``` 
We already need to start enforcing this anyway to avoid possible problems in production, so it goes in line with the current plan.

ivababukova (approved):

PR #488, opened 2023-09-28
[Cell-level metadata] Update the Add metadata select

StefanBabukov:
# Description

# Details
#### URL to issue
https://github.com/orgs/hms-dbmi-cellenics/projects/4/views/1?pane=issue&itemId=37286398
N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api.v2/controllers/cellLevelController.js
```diff
@@ -0,0 +1,43 @@
...
+
+const uploadCellLevelMetadata = async (req, res) => {
+  const { experimentId } = req.params;
+  const { name, createdAt } = req.body;
+
+  const cellLevelKey = uuidv4();
+  const bucketName = bucketNames.CELL_METADATA;
+  const newCellLevelFile = {
+    id: cellLevelKey,
+    created_at: createdAt,
```
cosa65: This one shouldn't be necessary, it's supped to update itself automatically

src/api.v2/controllers/cellLevelController.js
```diff
@@ -0,0 +1,43 @@
...
+    name,
+    upload_status: 'uploaded',
```
cosa65: Set it as `uploading`, not `uploaded`. It sohuld be uploaded when the upload is done

src/api.v2/controllers/cellLevelController.js
```diff
@@ -0,0 +1,43 @@
...
+  };
+  const cellLevelToExperimentMap = {
+    experiment_id: experimentId,
+    cell_metadata_file_id: cellLevelKey,
+  };
+
+  let uploadUrl;
+  await sqlClient.get().transaction(async (trx) => {
```
cosa65: Nice :)

src/api.v2/controllers/cellLevelController.js
```diff
@@ -0,0 +1,43 @@
...
+    await new CellLevel(trx).create(newCellLevelFile);
+    await new CellLevelToExperiment(trx).create(cellLevelToExperimentMap);
+
+    uploadUrl = await getSignedUrl('putObject',
+      {
+        Bucket: bucketName,
+        Key: cellLevelKey,
+      });
```
cosa65: I think this could be outside and not inside the transaction.

I think if the transaction fails the promise does too, so the whole function would fail and not generate the `uploadUrl`

src/api.v2/model/CellLevel.js
```diff
@@ -0,0 +1,40 @@
...
+];
+
+class CellLevel extends BasicModel {
+  constructor(sql = sqlClient.get()) {
+    super(sql, tableNames.CELL_LEVEL, fields);
+  }
+
+  async getMetadataByExperimentId(experimentId) {
+    try {
+      const result = await this.sql
```
cosa65: You can do `return await`

Also, the `try` `catch` doesn't really change much I think? The error is shown when it throws, right?

src/api.v2/model/Experiment.js
```diff
@@ -52,13 +53,6 @@ class Experiment extends BasicModel {
         ...aliasedExperimentFields,
         'm.key',
         'p.parent_experiment_id',
-        /*
-        The parent_experiment_id could be null in cases where the
-        parent experiment has been deleted.
-        The existence of a row with the experiment_id in the experiment_parent
-         table after doing a leftJoin,
-        indicates that it's a subsetted experiment.
-        */
```
cosa65: Why remove this?

cosa65 (commented):
There's some tests pending I think

src/api.v2/controllers/cellLevelController.js
StefanBabukov: Yep. I'll leave the handling of the upload status for the next ticket where the modal for the info will be implemented as well.

src/api.v2/controllers/cellLevelController.js
StefanBabukov: But keeping it inside is fine right? Seems the most readable way to know whats going on.

src/api.v2/model/CellLevel.js
StefanBabukov: didnt see any problems without having the try/catch for now so i'll remove it.

src/api.v2/controllers/cellLevelController.js
cosa65: As far as I understand, the transactions are supposed to encompass only the instructions that need to be in the transaction.

src/api.v2/model/Experiment.js
```diff
@@ -68,17 +69,26 @@ class Experiment extends BasicModel {
...
     );
 
-    return result;
+    // Assuming the getMetadataByExperimentId function is async and returns the metadata
+    const cellLevel = new CellLevel();
+
+    await Promise.all(experiments.map(async (experiment) => {
+      const result = await cellLevel.getMetadataByExperimentId(experiment.id);
+      experiment.cellLevelMetadata = result;
+    }));
```
cosa65: Why not get all the experiment's metadata at once? In the admin for example it's the difference between doing > 2000 queries vs only one.

You don't need to change the function `getMetadataByExperimentId` much, just pass as a param to this func the array of exp ids.

Then in the `where` map the array of experiment ids to `[{experiment_id: ...}, {experiment_id: ...}, ...]`

src/api.v2/controllers/cellLevelController.js
StefanBabukov: I was looking in other places where we have transactions and thats how it was done. For example in the `sampleFileController`

src/api.v2/model/Experiment.js
StefanBabukov: oh yeah, nice

ivababukova:
@StefanBabukov wouldn't it be better to use the name `cellLevelMeta` instead of `cellLevel`? We are shortening the name with 4 chars, but I feel like we loose a lot of descriptiveness and adding 4 more chars is worth it?

src/api.v2/controllers/cellLevelController.js
cosa65: oh, I feel like we could change those ones too then

src/api.v2/controllers/cellLevelController.js
```diff
@@ -0,0 +1,48 @@
...
+
+  res.json({ data: uploadUrlParams });
+};
+
+
+const updateCellLevelMetadata = async (req, res) => {
+  const { params: { experimentId }, body } = req;
+  const snakeCasedKeysToPatch = _.mapKeys(body, (_value, key) => _.snakeCase(key));
+  const { cellMetadataFileId } = await new CellLevelToExperiment().find({ experiment_id: experimentId }).first();
+  await new CellLevel().updateById(cellMetadataFileId, snakeCasedKeysToPatch);
```
cosa65: This call won't work, also we are allowing users to decide what to patch, and that's not something we want to do.

cosa65: This isn't part of the ticket anyways, so maybe you could just leave this function empty

src/api.v2/model/Experiment.js
```diff
@@ -68,17 +69,33 @@ class Experiment extends BasicModel {
...
+    const experiments = await collapseKeyIntoArray(
       mainQuery,
       [...fields, 'parent_experiment_id', 'is_subsetted'],
       'key',
       'metadataKeys',
       this.sql,
     );
 
-    return result;
+    // Assuming the getMetadataByExperimentId function is async and returns the metadata
```
cosa65: Do we need this comment? Could be removed I think

src/api.v2/model/Experiment.js
```diff
@@ -68,17 +69,33 @@ class Experiment extends BasicModel {
...
+    const cellLevel = new CellLevel();
+    const experimentIds = experiments.map((experiment) => experiment.id);
+    const cellLevelResults = await cellLevel.getMetadataByExperimentIds(experimentIds);
+
+    if (cellLevelResults.length) {
```
cosa65: This `if` shouldn't be necessary, if length === 0 then `forEach` doesn't do anything (there's nothing to iterate over).

src/api.v2/controllers/cellLevelController.js
StefanBabukov: I'm using this to update the status of the upload. What's the problem here? Why would it not work?

cosa65: I mean rather than snake casing anything that arrives, just extract the upload status and update only that.

It's more or less what the patch of sample files `const patchFile = async (req, res) => {` does

cosa65 (approved):
lgtm, I think there was one comment to address from iva too

PR #490, opened 2023-10-05
Fix upload of noncompressed files

cosa65:
# Description

The addition of multipart upload is not compatible with the pattern we had been following of making the create sample endpoint return the signed url necessary to upload the file itself.

This is because in its current state in `master`:
- To generate the signed urls to upload, we need to know how many parts the file is going to be split into --> **to create a file, we need to know its size**. As a consequence of this, when a file is not compressed, **we need to compress it before creating it** to store how its size will look like on creation.
- Apart from this, when a compression begins or fails, we need to update the file's upload status to display it to the user and to keep a record of it. To update a file's status, **we need to create a file before compressing it**.

Based on these points:
- we need to compress the file before creating it.
- we need to create the file before compressing it.

Clearly these 2 points don't work well together

### The fix 

There's one possible fix I discarded: mess with the upload status updates so that compression isn't shown or stored anywhere, I think this is likely to be a pretty complicated change to basically end up with less information and a degraded ux.

**The fix I went with** was to separate the sample files creation from the signed urls, now each of them have their own endpoint which are called by the ui, which works now in this order:
- create sample file.
- compress (with its upload status updates and all of that).
- get signed urls to multipart upload.

# Details
N/A https://ui-martinfosco-ui26-api10.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/905
https://github.com/biomage-org/ui/pull/26
https://github.com/hms-dbmi-cellenics/api/pull/490
https://github.com/biomage-org/api/pull/10

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/api.v2/controllers/sampleFileController.js
+++ b/src/api.v2/controllers/sampleFileController.js
@@ -51,7 +51,7 @@ const beginUpload = async (req, res) => {
   if (!['uploading', 'compressing'].includes(uploadStatus)) {
     throw new MethodNotAllowedError(
       `Sample file ${sampleFileId} is not in the process of being uploaded. 
-      No sample files can be replaced in s3, to replace a file referenced in a sample, create a new file for it`,
+      No sample files can be replaced in s3, to replace a file referenced by a sample, create a new file for it`,
     );
   }
 
--- a/src/api.v2/controllers/sampleFileController.js
+++ b/src/api.v2/controllers/sampleFileController.js
@@ -5,7 +5,7 @@ const SampleFile = require('../model/SampleFile');
 const bucketNames = require('../../config/bucketNames');
 
 const { getFileUploadUrls, getSampleFileDownloadUrl } = require('../helpers/s3/signedUrl');
-const { OK } = require('../../utils/responses');
+const { OK, MethodNotAllowedError } = require('../../utils/responses');
 const getLogger = require('../../utils/getLogger');
 
 const logger = getLogger('[SampleFileController] - ');
@@ -43,6 +43,17 @@ const beginUpload = async (req, res) => {
     params: { experimentId, sampleFileId },
     body: { metadata, size },
   } = req;
+
+  const { uploadStatus } = await new SampleFile().findById(sampleFileId).first();
+  // Check that the file is already in the process of being uploaded
+  // If it isn't, then it can't be reuploaded, a new file should be created
+  // This is because the existing files may be shared across many experiments
+  if (!['uploading', 'compressing'].includes(uploadStatus)) {
+    throw new MethodNotAllowedError(
+      `Sample file ${sampleFileId} is not in the process of being uploaded. 
+      No sample files can be replaced in s3, to replace a file referenced in a sample, create a new file for it`,
+    );
+  }
 
   logger.log(`Generating multipart upload urls for ${experimentId}, sample file ${sampleFileId}`);
   const uploadParams = await getFileUploadUrls(
--- a/src/api.v2/controllers/sampleFileController.js
+++ b/src/api.v2/controllers/sampleFileController.js
@@ -13,7 +13,7 @@ const logger = getLogger('[SampleFileController] - ');
 const createFile = async (req, res) => {
   const {
     params: { experimentId, sampleId, sampleFileType },
-    body: { sampleFileId, size, metadata = {} },
+    body: { sampleFileId, size },
   } = req;
   logger.log(`Creating file ${sampleFileType} for sample ${sampleId} in experiment ${experimentId}`);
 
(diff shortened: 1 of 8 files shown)


src/api.v2/controllers/sampleFileController.js
```diff
@@ -25,19 +25,31 @@ const createFile = async (req, res) => {
...
   });
 
 
   logger.log(`Finished creating sample file for experiment ${experimentId}, sample ${sampleId}, sampleFileType ${sampleFileType}`);
-  res.json(uploadUrlParams);
+
+  res.json(OK());
+};
+
+const beginUpload = async (req, res) => {
```
cosa65, line 41: The new endpoint to begin the upload.

src/specs/models/samples-bodies/CreateSampleFile.v2.yaml
```diff
@@ -6,15 +6,6 @@ properties:
     type: string
   size:
     type: number
-  metadata:
```
cosa65, line 9: Is used only in s3, so it is no longer necessary here

src/api.v2/controllers/sampleFileController.js
```diff
@@ -25,19 +25,42 @@ const createFile = async (req, res) => {
...
+  const { uploadStatus } = await new SampleFile().findById(sampleFileId).first();
+  // Check that the file is already in the process of being uploaded
+  // If it isn't, then it can't be reuploaded, a new file should be created
+  // This is because the existing files may be shared across many experiments
+  if (!['uploading', 'compressing'].includes(uploadStatus)) {
+    throw new MethodNotAllowedError(
+      `Sample file ${sampleFileId} is not in the process of being uploaded. 
+      No sample files can be replaced in s3, to replace a file referenced in a sample, create a new file for it`,
+    );
+  }
```
cosa65: The original motivation behind making the createFile endpoint return the signed urls was to make it harder for us to overwrite existing sample files which are shared across many different experiments. (you only get permission to upload to s3 when creating the file).

In order to maintain this security in some capacity, I added this check. Only files that are in the process of being uploaded (`'uploading'` or `'compressing'`) can have multipart uploads initiated for them

src/api.v2/controllers/sampleFileController.js
```diff
@@ -25,19 +25,42 @@ const createFile = async (req, res) => {
...
+  res.json(OK());
+};
+
+const beginUpload = async (req, res) => {
+  const {
+    params: { experimentId, sampleFileId },
+    body: { metadata, size },
+  } = req;
+
+  const { uploadStatus } = await new SampleFile().findById(sampleFileId).first();
```
kafkasl, line 47: just a random question, isn't it support to just pass the ID in the constructor? something like `new SampleFile(sampleFileId)` to make it less verbose? (no need to change here just to know)

src/api.v2/controllers/sampleFileController.js
kafkasl: can you explain a bit more why it's either `uploading` or `compressing`? I'm not familiar with the different status.

src/specs/models/samples-bodies/BeginSampleFileUpload.v2.yaml
```diff
@@ -0,0 +1,21 @@
+title: Begin sample file multipart upload
+description: 'Data to begin an s3 multipart upload'
+type: object
+properties:
+  size:
+    type: number
+  metadata:
+    description: Metadata to append to the s3 object on upload
```
kafkasl, line 8: the metadata is only cellranger version? what about the sample technologies that aren't 10x?

kafkasl (commented):
if I understood well, then currently uploads for big samples are broken?

src/api.v2/controllers/sampleFileController.js
cosa65: If you look at the possible statuses in the db you'll see these options:

'CREATE TYPE upload_status AS ENUM (\'invalidFormat\', \'uploaded\', \'uploading\', \'compressing\', \'uploadError\', \'fileNotFound\', \'fileReadError\', \'fileReadAborted\');',

Of these, only 2 of them imply that an upload is ongoing. That's why I chose only those 2

src/specs/models/samples-bodies/BeginSampleFileUpload.v2.yaml
cosa65: as far as I know 10x this is the only case where we use this s3 metadata thing.

I copied it over from the sample create validation where it was before without changing it so this is the same tha thas been around for over a year

src/api.v2/controllers/sampleFileController.js
cosa65: Not supported, the constructors can only receive transactions to run with for now.

src/api.v2/controllers/sampleFileController.js
kafkasl: I don't understand why compressing is considered upload, isn't that happening in the UI?

src/specs/models/samples-bodies/BeginSampleFileUpload.v2.yaml
kafkasl: ok

kafkasl (approved):

src/api.v2/controllers/sampleFileController.js
cosa65: it's considered part of the upload process, we could remove it if that's what you were thinking about (although not in this PR I think, partly because also removing it wouldn't get fully rid of this problem because we would still need to notify and store if the compression failed)

PR #491, opened 2023-10-11
scale replicas to 1 if set to zero

alexvpickering:
# Description
https://github.com/hms-dbmi-cellenics/api/pull/491

# Details
#### URL to issue
https://github.com/hms-dbmi-cellenics/issues/issues/77

N/A

#### Integration test branch
master

# Merge checklist

### Code updates
- [x] All new dependency licenses have been checked for compatibility

### Manual/unit testing

### Integration testing

### Documentation updates
- [x] Relevant Github READMEs updated **or** no GitHub README updates required.
- [x] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional


src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
```diff
@@ -62,7 +104,14 @@ const createWorkerResources = async (service) => {
     return { name, creationTimestamp, phase };
   }
 
-  const pods = await getAvailablePods(namespace);
+  // scale pods if worker has zero replicas
+  let pods = await getAvailablePods(namespace);
+  const { spec: { replicas } } = await getDeployment('worker', namespace);
```
ivababukova: wouldn't that line error out if `getDeployment` returns nothing?

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
```diff
@@ -62,7 +104,14 @@ const createWorkerResources = async (service) => {
...
+  if (pods.length < 1 && replicas === 0) {
```
ivababukova: I would move the `replicas === 0` check in the `scaleDeploymentReplicas` to remove the need to call function `getDeployment` twice. I will also modify the check to `replicas === desiredReplicas`, where `desiredReplicas` is the current third parameter of `scaleDeploymentReplicas`, called `replicas`, but renamed.

ivababukova: In this if statement: `if (pods.length < 1 && replicas === 0)` about this part: `pods.length < 1`. Is 1 the value of desired replicas? If so, I suggest creating a new variable called `desiredReplicas` that you can also pass in line 111 when you are calling function `scaleDeploymentReplicas`, so that we avoid hardcoding 1.

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
```diff
@@ -43,6 +46,45 @@ const getAvailablePods = async (namespace) => {
   return pods;
 };
 
+const getDeployment = async (name, namespace) => {
+  const k8sApi = kc.makeApiClient(k8s.AppsV1Api);
+
+  // find the particular deployment
+  const { body: deployment } = await k8sApi.readNamespacedDeployment(name, namespace);
```
ivababukova, line 52: In what cases can `k8sApi.readNamespacedDeployment` return an exception? I wonder if we aren't missing some exception handling here

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
```diff
@@ -43,6 +46,45 @@ const getAvailablePods = async (namespace) => {
...
+  if (pods.length > 0 || currentTry >= maxTries) {
+    return;
+  }
+
+  await waitForPods(namespace, maxTries, currentTry + 1);
+};
+
+
+//
+const scaleDeploymentReplicas = async (name, namespace, replicas) => {
```
ivababukova: rename parameter `replicas` to `desiredReplicas`, so that it is clearer

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
alexvpickering: I originally had the check in `scaleDeploymentReplicas` for the reason you mentioned but felt that `scaleDeploymentReplicas` should be explicitly for scaling the replicas as opposed to also checking if the condition is such that we want to scale. The idea being the it is slightly confusing to always call `scaleDeploymentReplicas` when in many cases it will not scale the replicas. Happy to change though if you think this is outweighed by the added simplicity or is not a concern.

For the check of `pods.length < 1`, 1 is not the number of desired replicas, it is the number of pods that needs to be available. You could change this check to `pods.length < minDesiredReplicas` but it is slightly confusing as they are different things we are checking (the number of available pods versus the value of `replicas` in the configuration). These should always have the same value in steady state but are different concepts.  In the absence of this PR, there is still a check that `pods.length < 1` in which case an error is thrown. This just adds one additional opportunity to not throw an error if the deployment is configured to have `replicas = 0`. If we want to remove the hardcoded `1`, we could have a `minAvailablePods` as well as a `minDesiredReplicas`.

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
alexvpickering: yes it would although so would e.g. `getPods` (called by `getAssignedPods` and `getAvailablePods`) . Not clear to me why we need be especially cautions about coding for an unhandled error in this case but not the other. If there is no deployment, there will also be no pods and so the failure would happen further up

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
alexvpickering: My guess would be if the `name` of the deployment ('worker' hardcoded) or `namespace` didn't exist or if the api didn't have permission

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
ivababukova: It is not like we need to be especially cautious for this case -- it is in general a best practice to handle possible error cases in the code.  We should have been cautious for all cases. If there is no error handling for some cases, where the code can fail, this it is something that needs to be corrected, not used as an example.

ivababukova: Of course, there are some limits to everything, but in my opinion in the api we definitely don't have error overhandling problem. In general, I don't think I've seen a piece of software that suffers from too much error handling :D

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
ivababukova: About this:

> I originally had the check in scaleDeploymentReplicas for the reason you mentioned but felt that scaleDeploymentReplicas should be explicitly for scaling the replicas as opposed to also checking if the condition is such that we want to scale. The idea being the it is slightly confusing to always call scaleDeploymentReplicas when in many cases it will not scale the replicas. Happy to change though if you think this is outweighed by the added simplicity or is not a concern.

I understand what you mean. Could we get the deployment once, as you are already getting it on line 108 (`const { spec: { replicas } } = await getDeployment('worker', namespace);`), and then instead of getting the deployment a second time in `scaleDeploymentReplicas`, pass it as a parameter to `scaleDeploymentReplicas`? My thinking here is that it avoids the possible confusion that you outlined, which is a good point, and saves 1 extra call to getDeployment.

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
ivababukova: If there is no deployment, wouldn't you want to raise an Exception here?

ivababukova: I am thinking from the perspective of the other environments (all non HMS productions) that don't require this scaling -- is there a possibility that `getDeployment` returns nothing, or returns an exception, but still we would want to proceed with the rest of the code?

There is handling, see line 115-117 in this file for example:
```
  if (pods.length < 1) {
    throw new Error(`Experiment ${experimentId} cannot be launched as there are no available workers.`);
  }
```

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
alexvpickering: Honestly it just makes things a mess when I try to do this and adds more lines than it eliminates. If you like this better I will go for it but it looks super ugly to me. If you have an alternative happy to do that also:

**deployment as parameter**:

```javascript
const scaleDeploymentReplicas = async (name, namespace, deployment, desiredReplicas = 1) => {
  const k8sApi = kc.makeApiClient(k8s.AppsV1Api);

  logger.log(`Scaling ${name} from ${deployment.spec.replicas} to ${desiredReplicas} replicas...`);

  // edit
  // eslint-disable-next-line no-param-reassign
  deployment.spec.replicas = desiredReplicas;

  // replace
  await k8sApi.replaceNamespacedDeployment(name, namespace, deployment);

  // wait until we have pods
  await waitForPods(namespace);
};


....
// scale pods if worker has zero replicas
  let replicas;
  let deployment;
  const minDesiredReplicas = 1;
  let pods = await getAvailablePods(namespace);

  try {
    deployment = await getDeployment('worker', namespace);
    replicas = deployment.spec.replicas;
  } catch (e) {
    logger.log('Getting worker deployment failed, ignoring...', e);
  }

  if (pods.length < 1 && replicas < minDesiredReplicas) {
    await scaleDeploymentReplicas('worker', namespace, deployment, minDesiredReplicas);
    pods = await getAvailablePods(namespace);
  }
  ...
```

**call getDeployment twice**:

```javascript

const scaleDeploymentReplicas = async (name, namespace, desiredReplicas = 1) => {
  const k8sApi = kc.makeApiClient(k8s.AppsV1Api);

  const deployment = await getDeployment(name, namespace);
  logger.log(`Scaling ${name} from ${deployment.spec.replicas} to ${desiredReplicas} replicas...`);

  // edit
  deployment.spec.replicas = desiredReplicas;

  // replace
  await k8sApi.replaceNamespacedDeployment(name, namespace, deployment);

  // wait until we have pods
  await waitForPods(namespace);
};

...
// scale pods if worker has zero replicas
  let replicas;
  const minDesiredReplicas = 1;
  let pods = await getAvailablePods(namespace);
  try {
    ({ spec: { replicas } } = await getDeployment('worker', namespace));
  } catch (e) {
    logger.log('Getting worker deployment failed, ignoring...', e);
  }

  if (pods.length < 1 && replicas < minDesiredReplicas) {
    await scaleDeploymentReplicas('worker', namespace, minDesiredReplicas);
    pods = await getAvailablePods(namespace);
  }
  ...
```

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
alexvpickering: I added error handling to ensure that `getDeployment` can fail and the function will still proceed

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
```diff
@@ -62,7 +102,21 @@ const createWorkerResources = async (service) => {
...
+  let pods = await getAvailablePods(namespace);
+  let replicas;
+  const minDesiredReplicas = 1;
+  try {
+    ({ spec: { replicas } } = await getDeployment('worker', namespace));
+  } catch (e) {
+    logger.log('Getting worker deployment failed, ignoring...', e);
+  }
+
+  if (pods.length < 1 && replicas < minDesiredReplicas) {
```
alexvpickering: Note: `replicas < minDesiredReplicas` evaluates to `false` if the above trycatch fails (in which case replicas is still undefined)

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
ivababukova: My understanding is that the code should rather look like this:

```
const scaleDeploymentReplicas = async (name, namespace, deployment, desiredReplicas = 1) => {
  const k8sApi = kc.makeApiClient(k8s.AppsV1Api);

  logger.log(`Scaling ${name} from ${deployment.spec.replicas} to ${desiredReplicas} replicas...`);

  // edit
  // eslint-disable-next-line no-param-reassign
  deployment.spec.replicas = desiredReplicas;

  // replace
  await k8sApi.replaceNamespacedDeployment(name, namespace, deployment);

  // wait until we have pods
  await waitForPods(namespace);
};

....
// scale pods if worker has zero replicas
  const minDesiredReplicas = 1;
  let pods = await getAvailablePods(namespace);

  try {
    const deployment = await getDeployment('worker', namespace);
    const replicas = deployment.spec.replicas;
    if (pods.length < 1 && replicas < minDesiredReplicas) {
      await scaleDeploymentReplicas('worker', namespace, deployment, minDesiredReplicas);
    }
  } catch (e) {
    logger.log('Getting worker deployment failed, ignoring...', e);
  }

   pods = await getAvailablePods(namespace);
  ...
```

I moved `pods = await getAvailablePods(namespace);` outside of the if statement, because it currently is executed like this, from what I can see in the changes diff. `if (pods.length < 1 && replicas < minDesiredReplicas)` should be in the try and catch, because it is coupled with getting `deployment` -- if `getDeployment` errors out or returns nothing, then even if the if statement is true, it will not execute with correct value of deployment. Am I missing something? What do you think about this code?

alexvpickering: It won't error because `replicas < minDesiredReplicas` will still evaluate to false as replicas has been instantiated (`let replicas;`). I like the above as it gets rid of some of this instantiation though. 

I've changed it to mostly reflect your suggestions (see below comment for one difference).

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
alexvpickering: For the `getAvailablePods` it only runs it once currently. The second call is directly tied to the scaling as it needs to be called again after scaling --> I think it makes sense to keep it inside the if statement

src/api.v2/helpers/worker/workSubmit/createWorkerK8s.js
alexvpickering: For the `getAvailablePods` it only runs it once currently. The second call is directly tied to the scaling as it needs to be called again after scaling --> I think it makes sense to keep it inside the if statement

ivababukova: my mistake on the `getAvailablePods` -- you are right

ivababukova: thanks for updating the code!

ivababukova (approved):

PR #494, opened 2023-10-25
Refactor etag

kafkasl:
# Description
The API now handles all the ETag generation and the dependencies of work requests. The socket use during work requests by the API has been removed.
* getExtraDependencies is in the API
* the api now responds directly to work requests returning  ETag & Signed URL (instead of using the socket) 
* ETag generation is fully in the API

TODO: 
* add the feature of returning non-cacheable ETags when running in development by adding a random small parameter.

# Details
#### URL to issue
https://github.com/hms-dbmi-cellenics/issues/issues/11

https://ui-ahriman-etag.scp-staging.biomage.net/data-management

https://github.com/hms-dbmi-cellenics/ui/pull/920
https://github.com/hms-dbmi-cellenics/api/pull/494

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/specs/models/WorkRequest.v2.yaml
```diff
@@ -4,15 +4,12 @@ description: This is the schema for the event emitted by the UI when certain wor
 properties:
   ETag:
     type: string
-    pattern: '^[a-fA-F0-9]+$'
+    pattern: "^[a-fA-F0-9]+$"
     description: A UUID generated by the client submitting the work. Will be used to map `WorkResponse` objects to the request submitted.
```
ivababukova, line 8: this should be deleted from the spec, because the UI no longer sends to Etag to the api, right?

ivababukova:
Did you notice that there are 2 functions, called `submitWork`: one in `submitWork.js` and one in `workController.js`?

src/api.v2/helpers/worker/generateEtag.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const { workerVersion } = config;
+
+  const taskName = body.name;
+  const extraDependencies = await getExtraDependencies(experimentId, taskName);
+
+  const ETagBody = {
+    experimentId,
+    body,
+    qcPipelineStartDate: qcPipelineStartDate.toISOString(),
```
ivababukova, line 29: When I run this code locally, I get an error on this line. The error I get is: `qcPipelineStartDate.toISOString is not a function`. The type of `qcPipelineStartDate` for me is a string and its value is `2023-10-05T11:51:48.691Z`. The problem must be that its type is not a date.

src/api.v2/helpers/roles.js
```diff
@@ -20,7 +20,7 @@ const allowedResources = {
     '/experiments/(?<experimentId>.*)/cellSets',
     '/experiments/(?<experimentId>.*)/clone',
     '/experiments/(?<experimentId>.*)/subset',
-    '/workRequest/(?<experimentId>.*)/.*',
```
ivababukova, line 23: why is the `/.*` removed?

src/api.v2/helpers/worker/workSubmit/getExtraDependencies.js
```diff
@@ -1,12 +1,48 @@
...
+      embeddingSettings: { method, methodSettings },
+    },
+  } = processingConfig;
+
+  return { method, methodSettings: methodSettings[method] };
 };
 
-const getCellSetsThatAffectDownsampling = async (message, body, cellSets) => {
+const getCellSets = async (experimentId) => {
+  // TODO consider just fetching latest modified date instead of the whole object
```
ivababukova: this TODO is complete, right? I see that the `getLastModified` function is returning the last modified date.

src/api.v2/helpers/worker/workSubmit/getExtraDependencies.js
```diff
@@ -1,12 +1,48 @@
...
+  const lastModified = await getLastModified({
+    Bucket: bucketNames.CELL_SETS,
+    Key: experimentId,
+  });
+
+  const lastVersion = `${bucketNames.CELL_SETS}/${experimentId}/${lastModified}`;
```
ivababukova, line 39: if ``${bucketNames.CELL_SETS}/${experimentId}` is empty, because there are no cell sets, then the value of `lastModified` will be null. Then the value of `lastVersion` will be either something invalid, or line 40: `const lastVersion = `${bucketNames.CELL_SETS}/${experimentId}/${lastModified};` will generate an error. Right? Am I missing something?

src/specs/api.v2.yaml
```diff
@@ -1432,8 +1432,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/HTTPError"
-  ? "/experiments/{experimentId}/samples/{sampleId}/metadataTracks/{metadataTrackKey}"
```
ivababukova, line 1435: these ? signs here and earlier are weird. Is this even syntactically correct?

src/specs/api.v2.yaml
```diff
@@ -2042,12 +2036,12 @@ paths:
...
-          content:
-            application/json:
-              schema:
-                $ref: "#/components/schemas/HTTPError"
+        # "404":
+        #   description: The work results were not found.
+        #   content:
+        #     application/json:
+        #       schema:
+        #         $ref: "#/components/schemas/HTTPError"
```
ivababukova: if this code is not needed, can you delete it?

src/specs/models/WorkRequest.v2.yaml
kafkasl: this is not correct because that `WorkRequest.v2.yaml` is the one the API sends to the worker. And the Etag is there (generated by the API). See `src/api.v2/events/handleWorkRequest.js:35`

kafkasl:
> Did you notice that there are 2 functions, called `submitWork`: one in `submitWork.js` and one in `workController.js`?

I didn't notice but I didn't introduce this in this PR so I leave for future refactors.

src/api.v2/helpers/roles.js
kafkasl: because it refers to the `ETag` which is no longer there.

src/api.v2/helpers/worker/workSubmit/getExtraDependencies.js
kafkasl: correct, removed comment.

src/api.v2/helpers/worker/workSubmit/getExtraDependencies.js
kafkasl: If there are no cellsets it will throw an error yes.  I think this is the desired behavior as you can't never send a work request unless the pipeline is done (and the cellsets are generated). I could do more defensive programming and return `lastModified=null` if the file does exist. It would make sense that, as long as it is still null, the dependency hasn't really been updated. I am not sure if I prefer that, or having the function fail fast because we assume there MUST be a cellset for work requests

src/specs/api.v2.yaml
kafkasl: no idea

kafkasl: I think it was removed by the formatter

src/specs/api.v2.yaml
kafkasl: done

src/api.v2/helpers/worker/generateEtag.js
kafkasl: I've just run it locally and in staging and work ok 
```
[2023-11-21T09:21:15.153Z] dependencies []
[2023-11-21T09:21:15.153Z] qcPipelineStartDate object 2023-11-21T09:16:36.676Z
[2023-11-21T09:21:15.153Z] ETagBody {
```
Maybe it was old data?

ivababukova: this must be out of date then

src/specs/models/WorkRequest.v2.yaml
ivababukova: makes sense

src/api.v2/helpers/worker/workSubmit/getExtraDependencies.js
ivababukova: I also think that it is better to raise an error and fail fast, rather than try to hide it by making `lastModified=null` only to have a failure later in the code. With the current version in the code, theoretically there is way in which `lastModified` is `null` or `undefined` without any error being generated. This will happen if `metadata.LastModified` in this line: https://github.com/hms-dbmi-cellenics/api/pull/494/files#diff-a48ec42db65954d588678dca5b1e781ee3e0c643ed3ebe9abadd0c45ec64998fR12 Is it possible for `metadata.LastModified` to return `null` or `undefined`?

src/api.v2/events/handleWorkRequest.js
```diff
@@ -1,62 +1,42 @@
...
+const generateETag = require('../helpers/worker/generateEtag');
 const getWorkResults = require('../helpers/worker/getWorkResults');
 
-const logger = getLogger();
 const validateAndSubmitWork = require('./validateAndSubmitWork');
 
-const handleWorkRequest = async (socket, data) => {
-  const { Authorization, experimentId, ETag } = data;
 
+//
```
cosa65: Remove

src/api.v2/events/handleWorkRequest.js
```diff
@@ -1,62 +1,42 @@
...
-    });
-
-    return;
+    return signedUrl;
   } catch (e) {
-    if (e.status === 404) {
-      console.log(`Work result ${ETag} not cached, submitting request`);
-    } else {
+    // 404 => indicates there are no work results present in S3 which is fine
+    // othere error => should be raised
```
cosa65: othere

src/api.v2/events/handleWorkRequest.js
```diff
@@ -1,62 +1,42 @@
...
 const getWorkResults = require('../helpers/worker/getWorkResults');
 
-const logger = getLogger();
 const validateAndSubmitWork = require('./validateAndSubmitWork');
 
-const handleWorkRequest = async (socket, data) => {
-  const { Authorization, experimentId, ETag } = data;
 
+//
+const getSignedUrlIfAvailable = async (experimentId, ETag) => {
```
cosa65, line 8: I saw there was a good explanation in the UI in which cases we wouldn't have anything to download, could you add it here too?

src/api.v2/helpers/pipeline/getPipelineStatus.js
```diff
@@ -311,14 +311,6 @@ const getPipelineStatus = async (experimentId, processName) => {
...
     response = buildResponse(processName, execution, shouldRerun, error, completedSteps);
   } catch (e) {
-    // if we get the execution does not exist it means we are using a pulled experiment so
-    // just return a mock sucess status
-    // TODO: state machines in production are deleted after 90 days, return a successful execution
-    // if the execution does not exist in production so the user will not be forced to re-run
-    // the pipeline losing annotations. This will be addressed checking if the
-    // processed files exist in S3 to avoid allowing users to move onwards when the pipeline was not
-    // actually run.
-
```
cosa65, line 321: Why remove this explanation?

src/api.v2/helpers/s3/getObject.js
```diff
@@ -24,4 +24,6 @@ const getObject = async (params) => {
   }
 };
 
+
+
```
cosa65: I think this isn't necessary

src/api.v2/helpers/s3/getLastModified.js
```diff
@@ -0,0 +1,28 @@
+const getS3Client = require('./getS3Client');
+const NotFoundError = require('../../../utils/responses/NotFoundError');
+
+const getLastModified = async (params) => {
```
cosa65, line 4: Why is this necessary vs just getting the object from s3?

src/api.v2/helpers/worker/generateEtag.js
```diff
@@ -0,0 +1,39 @@
...
+const createObjectHash = (object) => hash.MD5(object);
+
+const generateETag = async (
+  data,
+) => {
+  const {
+    experimentId,
+    body,
+    extras,
+    cacheUniquenessKey = null,
```
cosa65, line 15: This `cacheUniquenessKey` is never used, so I would say just remove it as a parameter in the function.

You could also remove it fully (from the `ETagbody` too), that would mean that all the work results in prod will be invalidated but that's not necessarily a problem, we do it from time to time.

For context, it is something that agi had added to test out things, the key needs to be set from the browser so it should either be reimplemented or removed

src/api.v2/events/handleWorkRequest.js
kafkasl: not sure which one you are referring to, the cases are already commented in the code

src/api.v2/helpers/pipeline/getPipelineStatus.js
kafkasl: because afaik this is completely outdated since you started storing the last state machines status in SQL.

src/api.v2/helpers/s3/getLastModified.js
kafkasl: because I don't want to get the whole object, the last modified is enough (head operation).

src/api.v2/helpers/worker/generateEtag.js
kafkasl: I have it in my list to reimplement in the future. Running locally now, the cache will still be used which is not the case before this PR. I just didn't want to make this PR even longer because it's a nice to have for development, but not affecting production

src/api.v2/helpers/worker/workSubmit/getExtraDependencies.js
kafkasl: I don't think so as it is a s3 attribute. If the file is not there it will fail

src/api.v2/helpers/pipeline/getPipelineStatus.js
cosa65: Oh right

src/api.v2/helpers/s3/getLastModified.js
cosa65: Right. Correct if I'm wrong, but based on what I've seen, `lastModified` is used to then get the last modified object from cell sets, right? My question was about trying to understand in which way this is different from just doing `getObject`

cosa65: Ah ok I understand now, I was confused because the function that was calling this is called `getCellSets`.

I would suggest changing the name for `getCellSets` to something like `getCellSetsVersion` or `getCellSetsLastVersionKey`

src/api.v2/helpers/worker/generateEtag.js
cosa65: Ok, could you make a comment here then that this param is not working currently and that there are plans to reimplement it?

kafkasl: the param does work if it's passed. but it's not being passed by caller functions

cosa65: ok I just mean something explaining that this is planning to be reimplemented, because if you are leaving this here on purpose and someone runs into it and notices that it isn't being used in the future they might just remove it as a "clean up" becuase they lack all of this context

cosa65 (approved):

PR #505, opened 2023-11-02
Make process project show up for cell level

cosa65:
# Description

# Details
#### URL to issue
https://github.com/hms-dbmi-cellenics/issues/issues/44

N/A
https://ui-martinfosco-ui47-api21-pip.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/ui/pull/927
https://github.com/hms-dbmi-cellenics/pipeline/pull/345

https://github.com/biomage-org/pipeline/pull/13
https://github.com/biomage-org/api/pull/21
https://github.com/biomage-org/ui/pull/47

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api.v2/helpers/pipeline/pipelineConstruct/index.js
```diff
@@ -163,6 +193,8 @@ const createQCPipeline = async (experimentId, processingConfigUpdates, authJWT,
     {
       state_machine_arn: stateMachineArn,
       execution_arn: executionArn,
+      last_pipeline_params: { cellMetadataId: currentCellMetadataId },
+      retry_params: { clusteringShouldRun },
```
cosa65, line 209: It is stored here because, if the `configureEmbedding` step fails, we need to know whether the last time it was run it was running clustering or not.

An alternative to this could be having `clusteringShouldRun` in the `processingConfig` instead. I am not completely sure about this option because the differences between this prop and the rest of the processing configs still holds, but if anyone thinks that would make more sense let me know. 

Important to note, it would be stored inside the processing config as a sort of "record" of the config of the previous attempted run, but not as something that can be changed by the user in any way. It is a property calculated based on other properties. 

I didn't store it in `last_pipeline_params` because that stores the values that, when changed, cause a pipeline to rerun. This property is different in that it is changed as a consequence of a pipeline running, it doesn't cause the pipeline reruns.

src/api.v2/helpers/pipeline/pipelineConstruct/qcHelpers.js
```diff
@@ -17,13 +20,46 @@ const qcStepsWithFilterSettings = [
...
+  experimentId, processingConfigUpdatedKeys, backendCompletedSteps, status,
+) => {
+  if (
+    processingConfigUpdatedKeys.length === 0
+    && await getCellLevelMetadataFileChanged(experimentId)
+  ) {
+    return filterToStepName.configureEmbedding;
+  }
+
+  if (processingConfigUpdatedKeys.length === 0 && status !== FAILED) {
```
StefanBabukov: How come status is needed now? In case of FAILED would it not go to retry directly using the `retry` endpoint? How will it reach this point?

src/api.v2/helpers/pipeline/pipelineConstruct/index.js
kafkasl: what is exactly `currentCellMetadatId` and what is used for?
regarding the `clusteringShouldRun`, I think it's fine although it's a bit hacky. Maybe we don't need to store that and if the embedding fails just re-run also the clustering. I don't think this fails often enough to have to keep track of it. I haven't run into any embeddings failure that I can remember of. Maybe @gerbeldo or @saracastel input can be useful here. I would probably just make users re-run both and avoid having to deal with this in the code.

ivababukova: About this: 

> Important to note, it would be stored inside the processing config as a sort of "record" of the config of the previous attempted run, but not as something that can be changed by the user in any way. It is a property calculated based on other properties.

Why would `clusteringShouldRun` be stored in the `processingConfig`? If this is the case, then it sounds like in both options we are storing it in the processing config, so what is the benefit if adding it to `retry_params`?

cosa65: > what is exactly currentCellMetadatId

`currentCellMetadataId` is the id of the currently uploaded cell level metadata file, it is used to know whether to put Process project of Go to data exploration in the button in data management in a similar way as it is done for gem2s for the same button.

> Maybe we don't need to store that and if the embedding fails just re-run also the clustering

I guess so, the problem with that is that we fall to the problem of going to all the trouble of not overwriting the clustering just to end up doing it when something in cell level fails. I personally think we should be able to rerun a failed pipeline with the expected params rather than defaulting to something else, my question was more along the lines of visibilizing this way of doing it vs an alternative.

src/api.v2/helpers/pipeline/pipelineConstruct/index.js
```diff
@@ -176,6 +205,8 @@ const createQCPipeline = async (experimentId, processingConfigUpdates, authJWT,
...
```
ivababukova, line 209: At the moment, in `master` branch do we use `retry_params` for anything in the qc pipeline? Or is it a new field that you are introducing?

src/api.v2/helpers/pipeline/shouldPipelineRerun.js
```diff
@@ -89,16 +91,33 @@ const getGem2sParams = async (experimentId, rawSamples = undefined) => {
...
-const shouldGem2sRerun = async (experimentId, pipelineType) => {
+const getQCParams = async (experimentId) => {
+  const metadataResult = (
+    await new CellLevelMeta().getMetadataByExperimentIds([experimentId])
+  )[0] || {};
+
+  const cellMetadataId = metadataResult.id || null;
+
+  return { cellMetadataId };
+};
```
ivababukova, line 101: this function contains basically the same functionality as the one I created and merged as part of this ticket: https://github.com/hms-dbmi-cellenics/issues/issues/38 I think you should pull master branch of API

src/api.v2/helpers/pipeline/pipelineConstruct/index.js
```diff
@@ -100,46 +118,57 @@ const getMetadataS3Path = async (experimentId) => {
...
+ * @param {*} authJWT
+ * @param {*} previousJobId
+ */
+const createQCPipeline = async (experimentId, processingConfigDiff, authJWT, previousJobId) => {
   logger.log(`createQCPipeline: fetch processing settings ${experimentId}`);
 
   const { processingConfig, samplesOrder } = await new Experiment().findById(experimentId).first();
+
+  const [cellLevelMetadata] = await new CellLevelMeta().getMetadataByExperimentIds([experimentId]);
+  const { id: currentCellMetadataId = null } = cellLevelMetadata || {};
```
ivababukova, line 135: pull in master branch -- this is already implemented

src/api.v2/helpers/pipeline/pipelineConstruct/index.js
cosa65: the field is used in gem2s and seurat in `master`

src/api.v2/helpers/pipeline/pipelineConstruct/index.js
ivababukova: use the return value of `getMetadataS3Path` function

src/api.v2/helpers/pipeline/pipelineConstruct/qcHelpers.js
cosa65: The `retry` endpoint is implemented only for gem2s and seurat, not qc. I did try to implement this in the retry endpoint in a previous iteration of this feature but abandoned it

StefanBabukov (approved):

PR #514, opened 2023-11-29
Caching hooks

kafkasl:
# Description
Fixes:
* local env not using caching by default
* the marker heatmap and embeddings being pre-computed by hooks after the pipeline.

Can now run cache locally by using `USE_CACHE=true make run`. 
Documentation updated

**How to test in staging**
1. Run a pipeline for the first time
2. Check that the worker runs embedding & marker heatmap after the pipeline automatically, and save the ETags
3. Go to data exploration and open the devTools. 
4. Check that the 2 ETags precomputed by the pipeline appear in the network requests
5. Alternatively, when going to data exploration, make sure there are no new embeddings & marker heatmap requests (only list genes)

# Details
#### URL to issue
N/A

https://ui-ahriman-api24.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/api.v2/helpers/worker/workSubmit/submitEmbeddingWork.js
```diff
@@ -1,27 +1,21 @@
-const submitWork = require('./submitWork');
+const submitWorkForHook = require('./submitWorkForHook');
 
 const submitEmbeddingWork = async (message) => {
   const {
     experimentId, input:
-    { authJWT, config: { embeddingSettings: { method, methodSettings, useSaved } } },
+    { authJWT, config: { embeddingSettings: { method, methodSettings } } },
```
ivababukova: why is `useSaved` removed?

ivababukova: If we no longer need to use useSaved, we should also remove it from the API schemas and also from the worker.
Also, be careful with removing `useSaved` -- it is a parameter used by Seurat upload feature (see Alex's explanation about what `useSaved` is used for: https://github.com/hms-dbmi-cellenics/api/pull/446#discussion_r1370504471)

kafkasl: when `useSaved=true` (only happens in seurat upload) we don't compute embeddings, instead we reuse the ones inside the object. So when this flag is set to true, I've changed the code so it does not precompute the embeddings neither.

src/api.v2/helpers/worker/workSubmit/submitMarkerHeatmapWork.js
```diff
@@ -1,9 +1,16 @@
...
+const getObject = require('../../s3/getObject');
 
 const submitMarkerHeatmapWork = async (message) => {
   const { experimentId, input: { authJWT } } = message;
 
+  const { cellSets } = JSON.parse(await getObject({
+    Bucket: bucketNames.CELL_SETS,
+    Key: experimentId,
+  }));
+
```
ivababukova, line 13: Do we need to get the cellsets and overwrite the value of `body.downsampleSetttings.cellSets` with them because `body.downsampleSettings.cellSets` are out of date? If so, why are they out of date? Why aren't we fetching the correct cell sets from the worker (I believe we were or still are fetching cell sets from the worker to aid computation of tasks).

src/api.v2/helpers/worker/workSubmit/submitWorkForHook.js
```diff
@@ -0,0 +1,33 @@
...
+const generateETag = require('../generateEtag');
+
+
+const submitWorkForHook = async (experimentId, authJWT, body) => {
+  const extras = undefined;
+
+  const data = {
+    experimentId,
+    body,
+    extras,
```
ivababukova: if `extras` is always undefined, why would we not remove it completely?

ivababukova:
I tried testing in staging, but when I am uploading files to my new project, I get upload error. Here are the logs from the API. Note that when I tried uploading files yesterday, it worked fine.

Screenshot 2023-12-01 at 11 29 38

src/api.v2/helpers/worker/workSubmit/submitMarkerHeatmapWork.js
kafkasl: we are not overwriting anything, this is a hook called only from the API and they need to be in the body only for caching purposes. In order to remove some of the weird `getExtraDependencies` parts, now the UI sends the exact name of all the cell sets (not only the hierarchy keys such as 'louvain'), so whenever there is any cell set included or removed the caching just works. In order to mimic that functionality from this hook, we need to manually add the default cell sets to the request.

src/api.v2/helpers/worker/workSubmit/submitWorkForHook.js
kafkasl: it is not needed indeed, removed

kafkasl:
> I tried testing in staging, but when I am uploading files to my new project, I get upload error. Here are the logs from the API. Note that when I tried uploading files yesterday, it worked fine.
> 
> Screenshot 2023-12-01 at 11 29 38

this is because the UI for the staging is not auto-updating, not sure why

src/api.v2/helpers/worker/workSubmit/submitMarkerHeatmapWork.js
ivababukova: thanks for the explanation, makes sense

ivababukova (approved):

